Issue #3, opened 2026-08-30
Services in Darnassus and Undercity may no longer exist


CybotTM:
`ServicePOIs.lua` advertises auction houses, banks, void storage and crafting tables in Darnassus (uiMapID 89) and Undercity (uiMapID 90). Both cities were destroyed in the Battle for Azeroth intro and have not been functioning capitals since patch 8.0.1.

**What is confirmed and what is not.** Both maps still exist — `C_Map.GetMapInfo` on a live 12.1.0.69497 client returns "Darnassus" (mapType 3, parent 12) and "Undercity" (mapType 3, parent 13). What could not be checked from outside the game is whether the service NPCs are still there and usable. That is the whole question, and the map existing does not answer it.

**Why this was not simply removed.** If the services are gone, QuickRoute routes players to vendors that are not there. If they are still reachable — Undercity in particular has been partially accessible at times — removing the entries takes away working functionality. Both mistakes are silent, so the entries stay until someone confirms in game.

**Affected rows.** `QuickRoute/Data/ServicePOIs.lua`, the Darnassus and Undercity entries in all four service lists (auction house, bank, void storage, crafting table).

**What would settle it.** Fly to each city, look for an auction house, a bank and a void storage. If they are gone, the entries are removed and the replacements — Alliance night elves use Stormwind, Forsaken use Orgrimmar — are what the router should already find on its own through the normal city nodes.

Found during the 12.1.0 maintenance pass. Deliberately not acted on: this is a gameplay fact, and guessing at it either way degrades the addon.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_019wx8ueHuqUidp5yybDQ2CN)_

CybotTM:
Evidence from the client's tables that narrows this, without answering it.

**Both maps are conditionally visible, and almost no zone map is.** Of the 348 zone-type `UiMap` rows, only **19** carry a `VisibilityPlayerConditionID`. Darnassus (89) and Undercity (90) are two of them — and Darnassus shares its exact condition, 68088, with Teldrassil (57).

The other members of that list are the tell:

| map | condition | what it is |
|---|---|---|
| 57 Teldrassil, 89 Darnassus | 68088 | burned in Battle for Azeroth |
| 90 Undercity | 62731 | destroyed in Battle for Azeroth |
[...]

Issue #2, opened 2026-08-30
Flight paths are missing from the travel graph


CybotTM:
The flight path network is not in the travel graph. Flight paths are the game's main overland transport, and QuickRoute routes as if they did not exist.

**Evidence.** `Core/TravelTime.lua:27` declares `flight = 0, -- Flight path (calculated separately)`, but `C_TaxiMap` does not appear anywhere in `QuickRoute/`. "Calculated separately" turns out to mean not at all. The speed constant is dead.

**What this costs.** Between two zones on the same continent with no portal, the router falls back to `EstimateSameContinentTravel`, a flat estimate, or to walking edges from `ZoneAdjacency`. A real player takes the nearest flight master. So the estimates are wrong in both directions: routes that a flight path would make quick are priced as long walks, and the step list never says "fly to X", which is what the player would actually do.

**What it needs.**

- Read the network with `C_TaxiMap.GetAllTaxiNodes(uiMapID)`, which yields node positions, names and whether the character has the node unlocked.
- Add taxi nodes as graph nodes and flight legs as edges. Flight time is roughly proportional to distance but not exactly, and the game does not expose a duration, so it needs a measured constant or a table.
- Only edges between nodes the character has actually unlocked may be used, which means the taxi data is per character and belongs with the other player-state scans in `PlayerInventory`, not in a static data file.
- The graph already grows to several hundred nodes with dungeon entrances alone; a few hundred taxi nodes on top makes the `BuildGraph` cost question real rather than theoretical.

**Sequencing.** This should wait for the `ConnectViaContinentRouting` indexing work, since adding taxi nodes to a build that already scans all nodes per node would multiply an existing cost rather than add a new one.

Found during the 12.1.0 maintenance pass. Not a regression — this has never been implemented.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_019wx8ueHuqUidp5yybDQ2CN)_

CybotTM:
Reopening: #18 was merged as *part* of this and GitHub closed the issue automatically. The flight network does not exist yet, so the issue is not done.

[...]

Issue #1, opened 2026-08-30
Mythic+ dungeon teleports are not modelled


CybotTM:
Mythic+ dungeon teleports are not modelled at all. A character who has earned them can reach those dungeon entrances with a single spell, and QuickRoute instead routes them across the continent on foot.

**Evidence.** `grep -riE 'mythic|keystone|challenge' QuickRoute/` returns nothing, and neither `Modules/DungeonData.lua` nor `Modules/DungeonPicker.lua` contains the string `teleport`. The dungeon picker knows where every entrance is and has no idea that the player may be able to teleport straight to it.

**Why it matters here more than for a generic teleport.** This addon already has a dungeon picker and static entrance coordinates for 213 instances. The teleports land exactly at those entrances, so the data the router needs on the destination side already exists — what is missing is the edge from the player to it. For anyone doing Mythic+ this is the difference between a one-click route and a wrong one.

**What it needs.**

- A table of dungeon-teleport spell IDs keyed by `journalInstanceID`, so the existing entrance data supplies the destination.
- Detection through `IsSpellKnown`, the same way `PlayerInventory:ScanSpells` already handles class and racial teleports.
- An edge from `PLAYER_NODE` to the dungeon node, weighted like other instant teleports plus `loadingScreenTime`.
- A new `QR.TeleportTypes` value, or reuse of `SPELL` with a marker — the panel groups by type, so this is a display decision as much as a data one.

**Open question.** The spell IDs change per season as dungeon rotations change, and there are dozens. They should be gathered from a live client rather than copied from a site: `/qrverify` already lists known spells that the data files do not recognise, which is one way to collect them per account.

Found during the 12.1.0 maintenance pass. Not a regression — this has never been implemented.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_019wx8ueHuqUidp5yybDQ2CN)_

PR #4, opened 2026-08-30
Maintenance: WoW 12.1.0 currency, zone graph corrections, CI and test hardening

CybotTM:
Brings QuickRoute up to WoW 12.1.0 and works through a full maintenance audit of the addon. 38 commits, each one self-contained with its reasoning in the message.

## Why it matters

The addon did not load on the live client. `## Interface:` was `120001` (patch 12.0.1, February), the live build is `12.1.0.69497`, so the AddOns list flagged QuickRoute out of date and skipped it unless the player had ticked "Load out of date AddOns". Everything below was invisible behind that.

Three defects would have made loading it worse than not loading it, so they are fixed first and the interface bump comes last:

- `SecureButtons.lua` called `QR:RegisterCombatCallback` at file scope, but that method is declared in `QuickRoute.lua`, which the TOC loads twenty-six lines later. The chunk aborted every login with a visible Lua error, and the leave-combat equipment restore was never registered.
- `CooldownTracker` type-checked `C_Spell.GetSpellCooldownDuration` against `"number"`. It returns a `DurationObject`, so the branch was dead and the function returned early reporting every spell teleport permanently off cooldown — mage portals, Death Gate, all of them offered while on cooldown, and routes built on that.
- The teleport panel's icon lookup fell through to the global `GetSpellInfo`, removed in 11.0.2, whenever spell data was not cached yet. That is the common state right after login, which is exactly when the panel first paints.

## The zone graph was on the wrong maps

Twelve zone nodes carried the uiMapID of something else entirely: `67` and `68` are Maraudon, `11` is Wailing Caverns, `261` is the Blood Furnace, `700` is Icecrown Citadel, `101` is the Outland continent, `876` is the Kul Tiras continent labelled "Zandalar", `875` is the Zandalar continent listed among its own zones, `433` is The Veiled Stair where Isle of Thunder was meant and `504` is Isle of Thunder where Isle of Giants was meant, and in `DungeonEntrances.lua` `25` is Hillsbrad Foothills where Burning Steppes was meant and `49` is Redridge Mountains where Eastern Plaguelands was meant. In each case the entry's neighbours, coordinates and travel times were right and only the key was wrong.

The Pandaria case is worth naming because it was not only a wrong label: the graph had built edges from the wrong identity. The Veiled Stair — a landlocked zone between Valley of the Four Winds and Kun-Lai Summit, and where `DungeonEntrances` puts the Terrace of Endless Spring — was modelled as an island reachable only by a 120-second boat from Townlong Steppes, so every route to that raid entrance was priced with a crossing that does not exist.

Most of it was invisible because the author had noticed and papered over it: each wrong node was joined to the correct one with a `0.001` second edge commented "same zone". Dijkstra treated Maraudon and Feralas as the same place for a thousandth of a second, and every route through them looked plausible.

`DungeonEntrances.lua` carried the same mistake in its zone table, and for four zones it defined both a wrong constant and an `_ALT` holding the right one — then used both, so Blackfathom Deeps, Uldaman, the Netherstorm dungeons and the Twilight Highlands raids were each registered on a map they are not on. Blackrock Mountain's six instances sat on Hillsbrad Foothills and Stratholme on Redridge Mountains, so a static-fallback route to Molten Core put its waypoint in the wrong zone.

Crucible of Storms was registered on the Kul Tiras *continent* map, so coordinates surveyed for a zone map landed on the continent and the entrance was never offered to a player standing in Stormsong Valley. Correcting it needed care: `942` already had a constant and an entrance block, and a second constant holding the same value would have silently replaced the first, since both form the same table key.

Twelve dungeon entrances were bound to the wrong Encounter Journal instance: The Nexus carried The Oculus's id, Atal'Dazar carried Kings' Rest's, and the Caverns of Time and Icecrown groups were rotated. Clicking "The Nexus" in the dungeon picker routed to The Oculus.

## What was verified against a live client

Every game-data change here was checked against a running 12.1.0.69497 client rather than inferred. That mattered: the audit's own proposals were wrong about five map IDs, and following them would have written Maraudon in as Feralas and Feralas as Tanaris — silently, because a routing graph does not complain about a plausible-looking wrong number.

What the client settled: 46 map identities, 21 Encounter Journal instances, and 5 spell IDs. Where the live client could not be consulted, the client's own `UiMap` and `JournalInstance` tables were: that is what settled the Pandaria and Kul Tiras corrections, and what showed the test fixture still carrying two of the very beliefs the data files had just shed. What it could not settle is filed rather than guessed — see the deliberately-not-done section.

## Also in here

- The Coiled Isle, Vaults of Atal'Utek, Val, Naigtal and the two player-housing zones, and eight dungeon and raid entrances from patches 11.1 through 12.1.0, none of which the addon knew about.
- The Delve-O-Bot 7001, and the observation behind it: `PlayerInventory:ScanToys` iterates the addon's own data table, so a teleport missing from it is invisible by construction and only surfaces when a person notices.
- The player node kept its walk edges after a zone change, so any route computed between arriving somewhere and the next bag update started with a walk back into the previous zone.
- Teleport edge weights froze at graph-build time, so "consider cooldowns" silently stopped working after the first build, in both directions.
- Continent routing overwrote measured in-city walk edges with a free `0.001` estimate, losing the whole final approach from the ETA.
- `C_Map.GetMapWorldSize` returns width **and** height. Only the width was read, so a north-south walk was priced by the map's east-west extent — wrong by the map's aspect ratio.
- Four Tol Barad teleports pointed at map `773`, an orphan UiMap with no continent and no adjacency, while the graph models Tol Barad as `244`. Such a node is reachable only through the flat 240-second cross-continent fallback and never links to anything physically beside it.
- Secure buttons were never returned when a window closed — 134 rows' worth on a populated teleport tab — draining the 60-button pool one ESC at a time, and were left visible on screen for whole fights.
- `BuildGraph` went from 20 ms to 12 ms by indexing nodes instead of scanning, with the built graph byte-identical before and after.
- Every pinned GitHub Action was on a runtime that is EOL or about to be; the release job interpolated the tag name straight into five `run:` blocks in the job holding the store API keys; both store uploads and the localization upload were downgraded to warnings, so a release could finish green having published nothing. The release now also refuses to zip a package that is missing a file its `.toc` references.

## Tests

`tests/` had never been linted by any gate and carried ten warnings, four of which were assertions that had quietly stopped being made. The test runner printed a warning for a file that failed to load and reported the suite green — which is how the SecureButtons load error survived.

The runner now fails on a load failure, CI lints `tests/` too, and every fix in this branch was verified by watching its guard fail first. Where a test stayed green through the injected defect it was rewritten rather than kept, and where a test asserted a value the client contradicts it was corrected rather than relaxed.

- 11504 → 11623 assertions, 0 failures
- Luacheck 1.2.0: 0 warnings, 0 errors across 70 files, `QuickRoute/` and `tests/`

## Review rounds

Eight rounds of independent review ran against this branch, each on a named commit and each re-verified here before anything was acted on. GitHub Copilot's review was unavailable throughout — the account is out of its monthly, account-wide quota — so the rounds were run by a separate reviewer and the merge rests on a documented self-review. Fifty-two findings were fixed across eight follow-up commits. The last three rounds found nothing in the code: what they found were release-note lines that claimed more than the branch does, and fixes that shipped without a test that reddens when they are reverted.

Seven of the forty-five were regressions introduced by an earlier round's own fixes. That is why the rounds were repeated rather than stopped after one, and the chain is worth reading as a whole:

- `MiniTeleportPanel` leaked three secure buttons per fight. Its release path skipped every `ReleaseButton` under lockdown and dropped the references anyway, and `PLAYER_REGEN_DISABLED` fires with lockdown already active — so that was the branch every combat start took.
- The fix for that was then disarmed by a second one: `OnHide` also fires when an ancestor is hidden (Alt+Z, a cinematic) and cleared the flag the enter-combat callback is gated on. `MainFrame` had been given that guard one round earlier; the module whose release hands the buttons back had not.
- Fixing the player node to rebuild its edges on movement destroyed its teleport edges. `ReconnectPlayerNode` keeps them on purpose — its docstring says so — and the very next call wrote walking estimates over them, because the graph holds one edge per node pair and the only overwrite guard protected walk edges from travel estimates, not teleport edges from anything. A mage in Elwynn Forest lost Teleport: Stormwind after a single step, and nothing brought it back: only bag, toy, spell, equipment and restriction events mark the graph dirty, never movement. Caught in the fourth round, four commits after it was introduced.
- The guard for that was then added at two call sites and tested at one. Removing the untested one alone left the suite green while reinstating the whole regression for any player standing in a city they can teleport to.
- `UI:OnCombatEnd` was made to re-render the route, then re-rendered it twice in the common case, because `MainFrame`'s callback is registered first and already rebuilds a window that combat hid.
- `CollectQuestBlocks` marked an erroring objective-tracker enumerator as a recognised shape, which made the caller hide every quest teleport button. The narrower fix was still wrong for a tracker with several modules: one working sibling re-armed the flag after another had failed.
- The packaging check added in the first round failed open — a `grep` that matched nothing left the counter at zero and let the zip proceed.

Three guards were widened from the specific to the general once a second member of their class turned up, and each widening found another member immediately. The list of maps that must never be used as a zone was eight hand-written entries; it is now every cosmic, world and continent map in the client's table, and applying it found a continent sitting in its own zone list. The rule against overwriting a measured edge was stated once and given to every writer, which found the one write in continent routing that had no check at all. The test fixture's map names were checked against the client and 21 were wrong — among them the two beliefs a fix commit had just deleted from the data files, still live in the fixture that validates them.

The rounds also measured what the commit messages and the release notes claimed. Eleven guards turned out to have no test that reddens on their removal, in code whose commit message said otherwise. Eight are now covered; three cannot be reached from a built graph and say so where they stand rather than being counted as coverage. Three release-note lines described defects that only ever existed on this branch and were removed; one count was wrong in both directions it could be read; one named the wrong patch; and one omitted the most visible of the four sites it summarised.

One of those guards is worth naming, because getting it wrong is easy: both waypoint confirmations had been switched from English literals to the translated template, and the test compared the printed message against the template's enUS text. On enUS those two are the same string, so the test would have passed for exactly the defect it existed to catch. It now swaps in a sentinel template instead, and was verified with a mutation whose literal is byte-identical to the enUS text.

Two reported findings were withdrawn by the reviewer after testing them, and one was checked and deliberately not changed: the item cooldown path has no secret-value guard where the spell path does, because `C_Container.GetItemCooldown` carries no such annotation.

## Deliberately not done

- Coordinates for the nine new dungeon entrances place them in the right zone, which is what decides whether a route is possible, but are not surveyed. The blocks say so.
- Nexus Point Xenas keeps coordinates byte-identical to Windrunner Spire's on a different map. One is a copy; which one cannot be told from here, so it is marked `SUSPECT` rather than replaced by a second guess.
- Silvermoon services and portals now point at map `2393`, where the player actually stands, but their coordinates were surveyed on the old `110` and are marked unverified.
- Zen Pilgrimage points at map `809`, which the client reports as Karazhan. That it is wrong is certain; what it should be is not: #5.
- Darnassus and Undercity services: #3.
- Mythic+ dungeon teleports (#1) and flight paths (#2) are absent as mechanisms, not as data. Both are features rather than maintenance.


CybotTM:
## Self-review

GitHub Copilot's review was unavailable for this pull request. The account is out of its monthly, account-wide review quota, which does not recover before the reset, so no bot review ran on any commit here. This comment records the review that took its place.

Self-review: 058204e862bd

**How it was reviewed.** Ten rounds of independent review ran against this branch, each against a named commit, each read and re-verified here before anything was acted on. Fifty-five findings were fixed across ten follow-up commits. Eight of those findings were regressions introduced by an earlier round's own fixes — which is the reason the rounds were repeated rather than stopped after one, and the reason the last four rounds exist at all.

**What the repetition caught.** The most serious defect on the branch was found in round four, four commits after it was introduced: rebuilding the player node's edges on movement destroyed its teleport edges, so a mage in Elwynn Forest lost Teleport: Stormwind after a single step and nothing brought it back — only bag, toy, spell, equipment and restriction events mark the graph dirty, never movement. It was confirmed by running the same probe in both worktrees: on `main` the route stays `total=8.001 teleport id=3561` before and after moving; on the branch it had become `total=30 walk id=nil`. The guard for it was then added at two call sites and tested at one, and round five caught that too.

**What was verified rather than argued.** Every game-data change was checked against a running 12.1.0.69497 client, or against the client's own `UiMap` and `JournalInstance` tables where the live client could not be consulted. That mattered: the original audit's proposals were wrong about five map IDs, and following them would have written Maraudon in as Feralas. Every fix was verified by watching its guard fail first — where a test stayed green through the injected defect it was rewritten, and eleven guards turned out to have no test that reddens on their removal despite a commit message saying otherwise. Eight of those are now covered; three cannot be reached from a built graph and say so in a comment where they stand rather than being counted as coverage.

**What was rejected.** Two reported findings were withdrawn by the reviewer after testing them. One was checked and deliberately not changed: the item cooldown path has no secret-value guard where the spell path does, because `C_Container.GetItemCooldown` carries no such annotation. Four release-note lines described defects that only ever existed between two commits on this branch and were removed or rewritten — telling users a bug was fixed that no release had is worse than saying nothing.

**What is deliberately not done** is listed in the pull request description, and each item that needs a decision rather than a patch has an issue: #5, #6, #7, #8, #9.

**Gates on the current head:** 11626 assertions, 0 failures; luacheck 1.2.0 clean across 70 files; both CI checks green. The suite passes in the runner's discovery order; that it fails in reverse is pre-existing on `main` and filed as #8.

PR #11, opened 2026-08-31
fix: restore the shared state test_poirouting.lua borrows

### LINKED ISSUE #8: Test suite only passes in one file order — three failures in reverse
The test suite passes in the order `tests/run_tests.lua` discovers files, and fails in another. Running the files in reverse order produces three failures, all in `test_mapteleportbutton.lua`, all `attempt to call method 'HookScript' (a nil value)` out of `MapTeleportButton:Initialize`.

This is pre-existing: `main` at `2d2cfa2` reproduces the identical three failures with the identical error, and neither `MapTeleportButton.lua` nor its test was touched by #4.

The cause is shared state that one test file leaves behind and another depends on — the same class of coupling that #4 had to fix inside `test_miniteleportpanel.lua`, where a helper counted the whole shared secure-button pool and the combat callbacks are registered globally and never cleared (`QuickRoute.lua`'s `combatCallbacks` is a file-local table that is appended to on every `RegisterCombat()` and has no reset). A suite that only passes in one order gives no signal when a test is added, moved or run alone.

Two things worth doing, in this order: make `MockWoW:Reset()` (or the runner between files) actually restore the state these tests share — the frame registry the `HookScript` failure comes from, and the combat-callback list — and then have CI run the suite a second time in reverse order, so the property is enforced rather than rediscovered.

Found during the review of #4, where it cost one round of false measurements because a scratch copy had retained the reversed order.

CybotTM:
Closes #8.

## The issue named the wrong cause

It pointed at `MockWoW:Reset()` not restoring the frame registry, and at `QuickRoute.lua`'s `combatCallbacks` growing without bound. Neither produces the failures. `MockWoW:Reset()` never touches `_G.WorldMapFrame`, and `combatCallbacks` — which really does grow from 9 to 64 across a suite run, because test files re-run module `Initialize()` — is not on the path to any of them.

The cause is one file borrowing something and not putting it back.

`tests/mock_wow_api.lua:553` defines `HookScript` as an own field on each frame — no metatable, no shared class — and `_G.WorldMapFrame` is created exactly once, at `:1773`. So the six `WorldMapFrame.HookScript = nil` lines in `tests/test_poirouting.lua`, all written as cleanup, restore nothing. They delete the method for the rest of the process, and `MapTeleportButton:Initialize` calls it at `MapTeleportButton.lua:432`.

Two files are enough to see it:

```
test_mapteleportbutton.lua then test_poirouting.lua   124 passed, 0 failed
test_poirouting.lua then test_mapteleportbutton.lua   121 passed, 3 failed
```

The three failures are verbatim the ones in the issue.

## And a second leak in the same file

It also leaves `QR.db.lastDestination` and `QR.db.destinationLocked` set, written by the real `POIRouting:RouteToMapPosition` and the real UI path that the tests drive on purpose. `UI:RefreshRoute` reads both, so a later file's tests route to a destination they never picked:

```
test_ui.lua then test_poirouting.lua   221 passed, 0 failed
test_poirouting.lua then test_ui.lua   217 passed, 4 failed
```

This one does not show up in plain reverse order — only in orders where this file happens to run before `test_ui.lua`, which is why the issue never saw it.

## The fix

Cleanup restores the original method instead of nil-ing it, and a file epilogue puts the three `QR.db` keys back. The one place that genuinely needs the method absent — the "falls back to SetScript when HookScript unavailable" test — keeps its `nil` and restores afterwards.

## Evidence

Thirty random file orders, same seed, before and after:

| | red runs | distinct signatures | failures |
|---|---|---|---|
| before | 19 of 30 | 5 | 93 |
| after | 0 of 30 | 0 | 0 |

The five signatures are exactly the two leaks: `MapTeleportButton` 57 times from the deleted method, and four `RefreshRoute` assertions from the db keys. Nothing else in the suite crosses a file boundary.

## Keeping it that way

The runner takes `QR_TEST_ORDER=discovery|reverse` and CI runs both, in `ci.yml` and in the release path. Discovery order now goes through `LC_ALL=C ls`, so it means byte order on every machine rather than the runner's locale collation — otherwise a red CI run cannot be reproduced locally.

**No shuffle mode.** A random seed in CI reddens `main` on a leak nobody introduced that day, and two fixed orders already fail on any state that crosses a file boundary. The permutation sweep above stays a scratch tool rather than a shipped one.

## Verified

- 11626 assertions, 0 failures, in both orders
- Luacheck 1.2.0: 0 warnings, 0 errors across 70 files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,8 +31,16 @@ jobs:
         run: luacheck QuickRoute/ tests/ --config .luacheckrc
 
   tests:
-    name: Tests
+    # Both file orders. The files share one mock and one addon namespace, so a
+    # file that borrows global state and does not put it back only breaks
+    # whatever runs after it -- which in discovery order was nothing, for as long
+    # as nobody added a file at the wrong position.
+    name: Tests (${{ matrix.order }} order)
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        order: [discovery, reverse]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
@@ -42,4 +50,6 @@ jobs:
           luaVersion: "5.1"
 
       - name: Run tests
+        env:
+          QR_TEST_ORDER: ${{ matrix.order }}
         run: lua tests/run_tests.lua
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,10 +11,18 @@ jobs:
   # a release published having passed the Lua suite and never having been
   # linted, which is the gate a tag most needs.
   tests:
-    name: Tests
+    # Both file orders. The files share one mock and one addon namespace, so a
+    # file that borrows global state and does not put it back only breaks
+    # whatever runs after it -- which in discovery order was nothing, for as long
+    # as nobody added a file at the wrong position.
+    name: Tests (${{ matrix.order }} order)
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        order: [discovery, reverse]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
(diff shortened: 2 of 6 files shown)


CybotTM:
Self-review: 90ab46a3e5c1177b88623f06095ebf5934b19a1a

PR #13, opened 2026-08-31
fix: make the last-resort node connection deterministic

### LINKED ISSUE #9: Strategy-4 fallback picks an arbitrary node while looking like it picks the cheapest
`PathCalculator:ConnectViaContinentRouting`'s strategy-4 fallback picks the node it connects to like this:

```lua
local bestNode, bestTime = nil, math_huge
for otherName, otherData in pairs(self.graph.nodes) do
    if otherName ~= nodeName and otherData.mapID
        and CanOverwriteWithTravel(self.graph, nodeName, otherName) then
        local baseTime = CROSS_CONTINENT_TIME
        if baseTime < bestTime then
            bestTime = baseTime
            bestNode = otherName
        end
    end
end
```

`baseTime` is the constant `CROSS_CONTINENT_TIME` on every iteration, so `baseTime < bestTime` is true exactly once — on the first eligible node — and false forever after. The variables are named as if a cheapest candidate were being chosen; what is actually chosen is whichever node `pairs()` happens to yield first, which is Lua hash order and therefore not stable across runs or between clients.

The branch is a last resort: it runs only when nothing else connected the node, so it decides how an otherwise isolated node reaches the rest of the graph. Getting an arbitrary one is not obviously wrong at a flat 300 seconds — every candidate costs the same — but the shape hides that no selection is happening, and a later edit that makes `baseTime` vary (per-continent times are already available through `QR.GetCrossContinentTravel`) would silently keep picking the first node instead of the cheapest.

Two ways to close it: compute a real per-candidate time and keep the comparison, or drop the pretence and take the first eligible node with a comment saying every candidate costs the same. The first is probably what was meant.

Pre-existing — the shape is identical on `main`; #4 only added the `CanOverwriteWithTravel` condition, because that branch was overwriting measured walk edges and teleport edges. Found during the review of #4.

CybotTM:
Closes #9.

The strategy-4 fallback read like a search — `bestNode`, `bestTime`, a `<` comparison — while the value compared was the constant `CROSS_CONTINENT_TIME`. So the condition was true exactly once, on the first eligible candidate, and the node connected was whichever name `pairs()` happened to yield. That is Lua hash order: not stable between runs or between clients, and not reproducible from a bug report.

## Which of the issue's two resolutions

The issue offered computing a real per-candidate time, or taking the first eligible with a comment. Neither, quite.

A real per-candidate time is not computable here. This branch runs only when the node has no continent for the hub pass to price against — that is *why* nothing else connected it — so there is nothing to compare and no cheapest to find. Every candidate genuinely costs the same.

But "first eligible" does not fix what the issue actually complains about. Its complaint is instability, and `pairs()` is exactly as unstable whether or not a comparison sits in front of it. So: collect the eligible names, sort, take the first. Determinism, not optimality, and the comment says which.

## Measured before changing anything

Instrumented over **612 graph builds** — every zone in `ZoneAdjacencies`, both factions, with and without class portals:

```
Graph builds:                    612
Strategy-4 fallback entered:       0
Strategy-4 fallback wrote:         0
```

It cannot be reached from any graph the addon builds today. That is worth knowing before touching it, and it is now recorded in the comment so the next person does not have to redo the measurement.

It is kept rather than deleted because the state it exists for — a node with no continent — is precisely what a data change reintroduces. That is not hypothetical: map `773` was in exactly that state until the 12.1.0 pass, four Tol Barad teleports pointing at an orphan UiMap with no continent and no adjacency.

## The guard is wired

Reverting the sort reddens the new test with `got: Zeta Outpost` against `expected: Alpha Landing` — so for that candidate set the hash order really does differ from sorted order, rather than agreeing with it by luck and leaving the assertion vacuous.

## Verified

- 11629 assertions, 0 failures
- Luacheck 1.2.0: 0 warnings, 0 errors across 70 files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Two test files left shared state behind, so the suite passed in one file order and failed in others: a mock frame method deleted rather than restored, and two saved-variables keys. CI now runs both orders.
+- The last-resort node connection in continent routing took whichever candidate Lua's hash order yielded, behind variables named as if it searched for the cheapest. It is deterministic now.
 
 ## [1.11.0] - 2026-08-31
 
--- a/QuickRoute/Core/PathCalculator.lua
+++ b/QuickRoute/Core/PathCalculator.lua
@@ -1106,29 +1106,45 @@ function PathCalculator:ConnectViaContinentRouting(nodeName, mapID, x, y)
             end
         end
 
-        -- Fallback: if no hub/city nodes found, connect to single best node.
+        -- Fallback: no hub or city was connected, so attach the node to one
+        -- other node rather than leaving it isolated.
+        --
+        -- Every candidate costs the same. This branch runs only when the node
+        -- has no continent the hub pass could price against, so there is
+        -- nothing to compare and no cheapest to find -- the previous shape
+        -- looked like a search (bestNode, bestTime, a < comparison) while the
+        -- compared value was the constant CROSS_CONTINENT_TIME, so it always
+        -- took whichever name pairs() happened to yield first. That is Lua hash
+        -- order: not stable between runs, and not reproducible from a bug
+        -- report. Sorting picks the same node every time; it is for
+        -- determinism, not for optimality.
+        --
         -- Candidates the overwrite rules refuse are skipped rather than
-        -- overwritten: this branch was the one write in the whole strategy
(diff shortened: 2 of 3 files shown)


CybotTM:
Self-review: 5891d38c990cfe91cdf587d3f6562b6ce3ecef09

PR #14, opened 2026-08-31
feat: make the dungeon picker reachable

### LINKED ISSUE #6: DungeonPicker is loaded but unreachable — nothing ever initializes it
`Modules/DungeonPicker.lua` is listed in `QuickRoute.toc` and loads, but nothing outside the file ever refers to it. `grep -rn "DungeonPicker" QuickRoute/` returns exactly two hits beyond the module itself: the `.toc` line, and a comment in `DestinationSearch.lua` that borrows its backdrop styling. `DungeonPicker:Initialize` is never called and the module registers no events of its own, so its combat callback at `DungeonPicker.lua:485` is never registered either.

The practical effect is that the dungeon picker cannot be opened at all, and every change made to it is untestable through the addon — including the `DUNGEON_NO_ENTRANCE_COORDS` handling added in `92fff62` during the 12.1.0 maintenance pass.

This is pre-existing, not introduced by #4, and it needs a decision rather than a patch: either wire it up (a slash command or an entry point from `DestinationSearch`, which is what appears to have replaced it) or remove the module and its `.toc` line. Filing it rather than picking one, since which of the two is right depends on whether the picker is still wanted alongside the destination search.

Found during the review of #4.

CybotTM:
Closes #6. You chose "wire it up" over removing it.

The module shipped in every release — the `.toc` loads it, and it has a full frame, row pool, tier headers, search box and combat handling — but nothing ever called its `Initialize` and nothing opened it. So every change made to it was untestable through the addon and invisible to players, including the `DUNGEON_NO_ENTRANCE_COORDS` handling added during the 12.1.0 pass.

It joins the init step list and gets `/qrdungeons`, matching how every other module registers its own command at file scope.

## What came with making it live

Activating dead code means owning its behaviour, so two things had to be fixed before this is honest:

**`Initialize` is now idempotent**, like every other module's. It calls `RegisterCombat`, which appends to a callback list that is never cleared, so a second call would run this module's combat handler twice per event. The guard was missing precisely because nothing had ever called it once.

**`OnHide` got the guard `MainFrame` and `MiniTeleportPanel` already carry.** `OnHide` fires when an ancestor is hidden too — Alt+Z, a cinematic — and clearing `isShowing` there leaves `Toggle` believing the picker is closed while it is on screen, so the next `/qrdungeons` re-shows an already-visible frame.

## Checked before wiring it up

The module holds no secure buttons at all, so nothing joins the 60-button pool and there is no leak to inherit — `grep` for `SecureButtons` in it returns nothing, and its `ReleaseAllRows` recycles plain frames. That was worth establishing first: the same session found `MiniTeleportPanel` leaking three buttons per fight through exactly that path.

## Verified

Each change reddens only its own test:

|---|---|
| drop it from the init step list | `is initialized during addon startup` |
| remove the `OnHide` guard | `an ancestor hide does not clear isShowing` |
| remove the slash command | `has a slash command that toggles it` |

11635 assertions, 0 failures. Luacheck 1.2.0: 0 warnings, 0 errors.


CybotTM:
Self-review: af1dc6d9234eac3efe2324d624e3122cefd5bf5f

PR #15, opened 2026-08-31
feat: dim teleport steps whose Use button combat is blocking

### LINKED ISSUE #7: UI.combatDisabledButtons is never populated, so the dimming it tracks does not exist
`UI.combatDisabledButtons` (`Modules/UI.lua:24`) is described as tracking Use buttons dimmed during combat. It is wiped in three places (`:681`, `:1195`, `:2024`) and iterated in one (`:2016`, inside `UI:OnCombatEnd`), but nothing ever inserts into it — all five references are accounted for by `grep -rn "combatDisabledButtons" QuickRoute/`. The restore loop therefore has nothing to restore, and no Use button is ever dimmed in the first place.

This is pre-existing and is not the same thing as the combat behaviour fixed in #4: that change made `UI:OnCombatEnd` re-render the route so its Use buttons are recreated after a fight, because `ConfigureStepUseButton` refuses to take a secure button under lockdown. The dimming mechanism is a separate, never-implemented idea sitting in the same function.

Two ways to close it, and the choice is a design decision rather than a fix: implement the dimming (a step whose teleport is unusable in combat gets `SetAlpha(0.5)` and is restored on the way out), or delete the field and the loop now that `OnCombatEnd` does something real. The field was left in place rather than removed as a drive-by during #4.

Found during the review of #4.

CybotTM:
Closes #7. You chose "implement the dimming" over deleting the field.

`UI.combatDisabledButtons` was wiped in three places and iterated in one, and nothing ever inserted into it. The dimming it was named for did not exist, so a teleport step rendered under lockdown looked exactly like a step whose teleport is unavailable — same text, no Use button, no signal which of the two it was.

The field is renamed to `combatDimmedSteps`, because what gets dimmed is the *step*, not a button. During combat there is no button to dim: `GetButton` refuses under lockdown and a secure frame cannot be configured there. That is precisely why the step needs a signal of its own, and it is the point I flagged when you picked this option.

## Three things that only showed up while building it

**The dim has to run last.** `CreateStepLabel` already writes the frame's alpha as the route-progress channel — `0.6` for a completed step — and that runs *after* `ConfigureStepUseButton`. Dimming from inside the button path was silently overwritten: the frame was recorded as dimmed while its alpha read `1`. Measured, not reasoned:

```
Schritt 1: frame.teleportID=140192 alpha=1
gedimmt registriert: 1
```

**Undimming is a re-render, not `SetAlpha(1.0)`.** Forcing full alpha would erase the `0.6` a completed step carries. `OnCombatEnd` already re-renders when a teleport step lacks its button — which is exactly the dimmed case.

**`ReleaseStepLabelFrame` now resets alpha.** Frames are pooled, and a dimmed one came back dim on a route rendered *out* of combat: the same wrong signal in the other direction.

## And one thing removed rather than added

A "was anything dimmed" branch in `OnCombatEnd`. Deleting it changed no test, because the existing condition already covers the case — a dimmed step is by definition a teleport step without a Use button. Rather than keep a branch no mutation can redden, it is gone, with a comment saying why so it is not reintroduced as if it were protective.

## Verified

|---|---|
| remove the dim | `a teleport step rendered in combat is dimmed` |
| remove the pool alpha reset | `a step frame does not come back from the pool dimmed` |

11634 assertions, 0 failures. Luacheck 1.2.0: 0 warnings, 0 errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Added
 - `/qrdungeons` opens the dungeon and raid picker. The module shipped in every release but nothing ever initialized it or opened it, so it was unreachable.
+- A teleport step rendered during combat is dimmed. It gets no Use button under lockdown, so it used to look identical to a step whose teleport is unavailable.
 
 ## [1.11.0] - 2026-08-31
 
--- a/QuickRoute/Modules/UI.lua
+++ b/QuickRoute/Modules/UI.lua
@@ -21,7 +21,7 @@ QR.UI = {
     frame = nil,
     stepLabels = {},
     stepLabelPool = {},  -- Pool of reusable step label frames
-    combatDisabledButtons = {},  -- Track buttons disabled during combat
+    combatDimmedSteps = {},  -- Step frames dimmed because combat blocks their teleport
     -- Cache tables for localized names (reduces repeated API calls)
     itemInfoCache = {},
     spellInfoCache = {},
@@ -36,6 +36,9 @@ QR.UI = {
 }
 
 local UI = QR.UI
+
+-- Alpha for a teleport step whose Use button combat is blocking
+local COMBAT_DIMMED_ALPHA = 0.5
 
 -- Localization shorthand
 local L = QR.L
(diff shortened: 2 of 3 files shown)


CybotTM:
Self-review: fdf53ccee09f3b29c6d7114dbc716028c422e338

Issue #5, opened 2026-08-30
Zen Pilgrimage points at map 809, which the client reports as Karazhan


CybotTM:
`QuickRoute/Data/TeleportItems.lua` gives the Monk teleport Zen Pilgrimage the destination map `809`. The client's UiMap table calls 809 **Karazhan** (Type 4, parent 42), so the map is certainly wrong: a route to the Monk destination sets its waypoint in Karazhan.

What it should be is not settled. The entry's own destination string is "Peak of Serenity / Temple of Five Dawns", which are two different places, and the UiMap table has no map by either name — the Peak of Serenity is a subzone of Kun-Lai Summit (`379`) and the Temple of Five Dawns sits on The Wandering Isle (`2366`, with orphan variants `378` and `709`). Which one the spell lands on in 12.1.0 can only be decided by casting it on a live client and reading `C_Map.GetBestMapForUnit("player")`.

Until then the entry is listed in the `OFF_GRAPH` allow-list of `tests/test_data_validation.lua` ("Teleports: every destination map belongs to a continent") so that a *new* orphan destination still fails the suite while this known-wrong one does not. Closing this issue means: cast Zen Pilgrimage, record the map, correct the entry and its coordinates, and remove the 809 line from that allow-list.

Found while working through the review of #4, which corrected the same class of error for Silvermoon City (110 → 2393) and Tol Barad (orphan map 773 → 244).

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_019wx8ueHuqUidp5yybDQ2CN)_

CybotTM:
Two things established from the client's own tables, so nobody has to re-derive them.

**The reported defect is real.** `UiMap` row 809 is `Karazhan`, `Type = 4` (an instance floor), `ParentUiMapID = 42` (Deadwind Pass). `TeleportItems.lua:1104` points Zen Pilgrimage at it with the comment `-- Peak of Serenity`. Whatever the correct destination is, it is not that.

**The correct destination cannot be derived offline, and now there is a precise reason.** `SpellEffect` for spell 126892 has four rows; the teleport one carries `ImplicitTarget_1 = 17`, which is `TARGET_DEST_DB` — "the destination is a row in `SpellTargetPosition`". That is the one table the wago.tools exports do not publish: every other table this addon's data was built from downloads fine, and `SpellTargetPosition.csv` returns 404.

[...]

PR #16, opened 2026-08-31
feat: add /qrverifymap for the questions data cannot answer

CybotTM:
Unblocks #5 and #3. Neither closes with this PR — both close when someone runs the command in game and reports what it says.

## Why a command rather than a fix

Two open issues cannot be settled from the repository, and both were costing a round trip per question.

**#5** asks which map Zen Pilgrimage lands on. The entry says `809`, which the client calls Karazhan, and the destination string names two different places — "Peak of Serenity / Temple of Five Dawns", a subzone of Kun-Lai Summit and a spot on The Wandering Isle. That it is wrong is certain; what it should be is not derivable from any table in the repo.

**#3** asks whether the Darnassus and Undercity services still exist. Whether an auctioneer stands at a coordinate today is not recorded in `UiMap`, `AreaPOI`, or anything else available here.

## What it does

`/qrverifymap [mapID]` prints the client's view of a map and the addon's belief about it side by side, plus everything in the data that points at that map, into the same copyable window the other diagnostics use.

For `809` it puts the whole contradiction on one screen:

```
Reporting map 809 on request. The player is on 84.
client   : 809 Karazhan (type 3)
addon    : continent nil, 0 adjacent zone(s)
graph    : 0 node(s)

The addon points these at map 809:
  ClassTeleportSpells[126892] Zen Pilgrimage -> Peak of Serenity / Temple of Five Dawns
```

And for a map that is fine, the same shape reads as agreement:

```
The client places the player on map 84.
client   : 84 Stormwind City (type 3)  <-  13 Eastern Kingdoms (type 2)
position : 0.5000, 0.5000
addon    : continent EASTERN_KINGDOMS, 1 adjacent zone(s)
graph    : 12 node(s) -- Deeprun Tram (Stormwind to Ironforge) (Start), …
```

The parent chain is walked rather than just the name, because a wrong map is usually recognisable by its parent — that is how the Kul Tiras and Quel'Thalas corrections were settled during the 12.1.0 pass.

## What to do with it

For **#5**: on a Monk, cast Zen Pilgrimage, then `/qrverifymap`. The first line names the map you landed on.

For **#3**: stand in Undercity, `/qrverifymap`; then in Darnassus, `/qrverifymap`. The report names the map the client actually places you on and lists every service the addon has registered there, so "the addon sends people to an auctioneer that is not there" becomes checkable against what you can see.

## Verified

11637 assertions, 0 failures. Luacheck 1.2.0: 0 warnings, 0 errors.


CybotTM:
Self-review: fbd6b9a4350ad76d8f3450bc7318981f87cc491f

PR #17, opened 2026-08-31
feat: route through the dungeon teleports the player knows

### LINKED ISSUE #1: Mythic+ dungeon teleports are not modelled

CybotTM:
Closes #1.

A character holding the Mythic+ teleport for a dungeon was routed there on foot. The addon already knew the entrance — `AddDungeonNodes` builds a node per instance — and already scanned spells. Nothing connected the two.

## The data

Derived from the client's own tables rather than hand-listed: every `SpellName` row named `Teleport: <X>` where `X` is a `JournalInstance` name. **102 spells across 83 instances.**

That match is deliberately generous. It catches the Mythic+ rewards this issue is about *and* the older attunement teleports, and it cannot be narrowed from outside the game: no client table marks which is which, and `SpellCooldowns` records nothing for any of them. Fourteen instances have more than one candidate — `Halls of Valor` has four — and both are kept, since they share the `journalInstanceID` and only one can be learned.

**The generosity is safe because `IsSpellKnown` gates it.** A spell nobody can learn never becomes an edge, and neither does a boss ability of the same name.

That is not an argument, it is measured. Removing the gate reddens an *unrelated existing* route test:

```
CalculatePath: a destination inside a capital keeps its real legs:
  crossing between two capitals is not priced under a minute (got 25s)
```

A teleport the player does not own would shorten routes it has nothing to do with, and the suite already catches that.

## Why a separate build step

`AddDungeonTeleportEdges` is not part of `AddPlayerTeleportEdges`, for two reasons that are easy to get wrong:

- That one runs **before** `AddDungeonNodes`, so the nodes these edges point at do not exist yet. Moving the new step above `AddDungeonNodes` reddens the test.
- It builds its destination node from `data.mapID`, which these spells do not carry. A dungeon teleport lands at an entrance the graph already models, so it reuses that node rather than inventing a second one beside it.

No marker field on the scanned spell either: `GetAllTeleports` copies only `isClass` and `isMageTeleport`, so an `isDungeonTeleport` flag would be dropped before anything read it. The step keys on `data.journalInstanceID`, which is what actually distinguishes these.

## What is not settled

Whether every dungeon teleport follows the `Teleport: <instance>` convention. If one does not, it is invisible to this derivation and to any test built on it, and there is no way to enumerate the true set from outside the game. A wrong entry costs a teleport offered to a dungeon it does not lead to; `/qrscan` lists what was detected and `/qrverifymap` (#16) can be pointed at the destination.

## Verified

|---|---|
| remove the build step | `and the player has an edge to it` |
| remove the `IsSpellKnown` gate | the capital-crossing route test, plus the spell-ID assertion |
| move the step above `AddDungeonNodes` | `and the player has an edge to it` |

11632 assertions, 0 failures. Luacheck 1.2.0: 0 warnings, 0 errors across 71 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - `/qrdungeons` opens the dungeon and raid picker. The module shipped in every release but nothing ever initialized it or opened it, so it was unreachable.
 - A teleport step rendered during combat is dimmed. It gets no Use button under lockdown, so it used to look identical to a step whose teleport is unavailable.
 - `/qrverifymap [mapID]` reports what the client says about a map and what the addon believes about it, side by side, in the copyable window. For the questions that cannot be answered from the data — which map a teleport actually lands on, whether a service is still there.
+- Dungeon and raid teleports the player knows are now routes. A character with the Mythic+ teleport for a dungeon was sent there on foot: the addon knew the entrance and knew the spell, and nothing connected the two.
 
 ## [1.11.0] - 2026-08-31
 
--- a/QuickRoute/Core/PathCalculator.lua
+++ b/QuickRoute/Core/PathCalculator.lua
@@ -211,6 +211,17 @@ function PathCalculator:BuildGraph()
         buildError = buildError or err
     end
 
+    -- Dungeon teleports the player knows. Must come after AddDungeonNodes:
+    -- the nodes these edges point at do not exist before it.
+    success, err = pcall(function()
+        self:AddDungeonTeleportEdges()
+    end)
+    if not success then
+        QR:Error("AddDungeonTeleportEdges failed: " .. tostring(err))
+        buildSuccess = false
+        buildError = buildError or err
+    end
+
     -- IMPORTANT: Connect all nodes on the same map with walking edges
     -- This ensures teleport destinations connect to portal hubs on the same map
     success, err = pcall(function()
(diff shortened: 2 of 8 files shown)


CybotTM:
Self-review: 43de520422d7021b46fa6a4be7e0cfc7c609874b

PR #18, opened 2026-08-31
fix: connect the player node when the graph is built

CybotTM:
Part of #2. It does **not** close it — the flight network itself still needs a live client, and the reason is below.

## The defect underneath the issue

The issue is about flight paths, but the symptom it names — a teleport-less character stranded on several continents — has a cause that has nothing to do with flight.

Nothing connected the player node at build time except the teleports they own:

- `ConnectSameMapNodes` reaches it only when another node shares its map
- `ConnectIslandNodes` skips it by name (`if nodeName ~= PLAYER_NODE`)
- `ReconnectPlayerNode`, the only other caller of `ConnectNearbyNodes` for the player, is gated on the position having changed

So a character with no teleports, standing somewhere the graph has no other node for, had a player node with **zero edges** and no route anywhere. Thunder Bluff is the plainest case: one node on that map, the player's own.

## Measured, not deduced

A teleport-less warrior swept through all 153 zones in `ZoneAdjacencies`, routing to Stormwind:

| | zones with no route |
|---|---|
| before | **29** |
| after | **21** |

Neutral where edges already existed — 459 route comparisons for a mage with three teleports, across every zone and three destinations:

```
neu erreichbar : 0
schneller      : 0
unveraendert   : 459
langsamer      : 0
Route verloren : 0
```

## What this does not fix

The remaining 21 are all of Pandaria and Draenor. Those continents have portals *into* them in `QR.PortalHubs` and no modelled return leg, so a character without teleports has no way out in the data.

Adding those return portals needs coordinates I cannot verify from outside the game. Inventing them is the exact failure mode this codebase has been paying for: a plausible wrong number that a routing graph does not complain about — the 12.1.0 pass spent ten review rounds removing them.

Flight paths would solve it properly, which is what #2 is for. That work still needs a live client:

- no API returns a flight duration, so every edge weight would be a guess
- the real per-character flight-point count is unknown here; every number I could produce is synthetic
- whether `GetTaxiNodesForMap` on a continent map returns the union of its zones' nodes is unverified

## Verified

Removing `ConnectPlayerNode` reddens the new test with `the player node has edges straight after the build (got 0)`.

11629 assertions, 0 failures. Luacheck 1.2.0: 0 warnings, 0 errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 - Two test files left shared state behind, so the suite passed in one file order and failed in others: a mock frame method deleted rather than restored, and two saved-variables keys. CI now runs both orders.
 - The last-resort node connection in continent routing took whichever candidate Lua's hash order yielded, behind variables named as if it searched for the cheapest. It is deterministic now.
+- A character with no teleports had no route at all from anywhere the graph has no other node for -- Thunder Bluff, Darnassus, Ashran and 26 other zones. The player node was only connected by the teleports they owned, and by nothing else until they moved.
 
 ### Added
 - `/qrdungeons` opens the dungeon and raid picker. The module shipped in every release but nothing ever initialized it or opened it, so it was unreachable.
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review: 0b812a66980f631843946dfffb2a9b025d44598f

PR #20, opened 2026-08-31
feat: model the flight network

### LINKED ISSUE #2: Flight paths are missing from the travel graph

CybotTM:
Closes #2.

## What this does

Adds the flight network to the travel graph: 141 zones with a flight master, positioned from the client's own tables, connected where the player has actually discovered them, with each flight step's waypoint on the flight master itself.

## I was wrong twice, and both are worth reading before the diff

**First: I said this needed a live client.** In #18 and in a comment on this issue I wrote that no API returns a flight duration, so every edge weight would be a guess. `TaxiNodes` gives 1464 taxi nodes with world positions, `TaxiPath` gives the 8690-edge network between them, and `UiMapAssignment` converts a world position into a zone. The distance an edge is priced from is exact; only the speed it is divided by is estimated, and that is the same kind of constant as `SPEEDS.mounted_flying`, which this addon has always carried.

**Second: the first version of this PR was measured wrong.** A review of `06aa94e` found eight defects. Five were real, and two of them invalidated numbers I had put in the PR body:

- The headline "The Hinterlands 510s → 226s" was **produced by a bug**. Flight edges could anchor on the player node, and The Hinterlands has no other graph node, so that improvement existed only because the edge was attached to a position that stops being true the moment the player moves. It is not in the corrected numbers at all.
- "None got worse" was true of the nine zones I sampled and **false of the graph**. A one-directional portal was being replaced by a bidirectional flight write: Valdrakken → The Waking Shores went from a 10s portal to a 101s flight, `portalData` and all.

The tests I shipped were why this got as far as a ready-for-review PR. Five defect injections against them — dropping the teleport guard, always overwriting, inverting the continent check, ignoring the discovered set, dropping the determinism sort — all left 11671 assertions passing. Exactly one assertion was wired to the feature.

## What is fixed in `997342b`

| defect | fix |
|---|---|
| Flight edge could anchor on the player node and survive the player moving | `FlightAnchorForMap` excludes it |
| One-directional portal replaced by a bidirectional write | each direction decided on its own |
| `continentID` is a world map, not a taxi network — map 530 mixes Outland with the BC starting zones, giving 50 impossible flights | the addon's own continent must agree |
| `NodesOnMap[1]` is `pairs()` hash order; an unrelated teleport re-anchored six maps | sorted, zone nodes before dungeon nodes |
| `GetKnownFlightZones` could never return the empty table its own doc comment describes | it can now |
| A flight leg read "Go to X" everywhere but the route list | `BuildSteps` has a flight branch, reusing the existing `ACTION_FLY_TO` |

## Two things were actually wrong, not merely untested

Round 7 found both, and they turned out to be the same story.

**The waypoint pointed at the wrong building.** A flight step navigated to the graph node the edge hangs off — whatever `FlightAnchorForMap` ranked highest on that map. For the Badlands that is the Dragon's Mouth dungeon entrance, **0.48 of the zone away** from Fuselight. The addon told the player to fly, and sent them somewhere there is no flight master.

The first attempt at fixing this **made it worse**, and round 8 caught it: `AddFlightEdges` handed one `data` table to both directions of a pair, so on the reverse half `fromMapID` named the map the flight *arrives* at. **301 of 599 edges** carried it — just over half. Reading that field turned "right map, wrong building" into "wrong map entirely" — a player in Mount Hyjal sent to Teldrassil to board a flight leaving Mount Hyjal. Fixed at the source: each direction gets its own data. The test that had "verified" the first attempt walked one route whose single flight leg was the forward direction, so it was blind by construction; two tests replace it, one walking every edge in the graph and one routing out of Mount Hyjal.

The position it needed was in the data file and **read by no addon code at all**. `QR.FlightPoints[…].x/.y` — the coordinates the round-4 fix below was entirely about — were used by nothing, which is precisely why a transposed projection survived four rounds. The flight branch reads them now.

**Seven zones with real flight masters were excluded.** `related()` required a prefix match, so it could not see past a leading or trailing qualifier: `Camp Antonidas, Azure Span` against *The* Azure Span, `Nordrassil, Hyjal` against *Mount* Hyjal, `Booty Bay, Stranglethorn` against *The Cape of* Stranglethorn. All seven were discarded as "name contradicts geometry" while their geometry was right — including Emerald Dream and The Azure Span, current content.

Substring instead of prefix accepts those seven and still rejects every border case the rule exists for (Badlands for Searing Gorge, Duskwood for Deadwind Pass, Dalaran for Broken Shore, Southern Barrens for Mulgore, Northern Barrens for Durotar, Blasted Lands for Swamp of Sorrows). Regenerating: **+7, −0, ~0**. Mount Hyjal promptly shows up in the routing table at 705s → 418s.

## The zone coordinates were transposed

Separately from which zones are in the file: every `x`/`y` in it was wrong from the first commit until `771b9c6`. The UI axes are the transpose of the world axes and both run backwards — a zone's `x` comes from world Y, its `y` from world X — and the generator read them straight through, shipping `(1 - y, 1 - x)`.

`UiMapAssignment` says so. On uiMap 138 a base row plus two sub-boxes fit a single affine map only under that reading; on uiMap 13 two rows share the world-Y span `[-1400,-800]` and carry identical `UiMin_0/UiMax_0` while differing in world X and in `UiMin_1/UiMax_1`, which is only possible if `ui_0` is a function of world Y. Across the 133 multi-row uiMap groups, a good number are internally consistent *only* under this reading and **zero** only under any other — that asymmetry is the robust part; the exact count moves with the comparison tolerance, so it is not quoted here.

Stormwind's flight master shipped at `(0.2703, 0.2902)`. It is at `(0.7098, 0.7297)` — 918 yards apart; worst case was Zul'Aman at 4090.

## Data regenerated, 166 → 141 zones

`TaxiNodes` holds far more than flight masters, and the first version took whatever the geometry said. Five filters now, in order:

1. **Not tagged internal or disabled by name** — `[Hidden]`, `[DISABLED in 4.x]`, `[TEMP]`.
2. **At least two `TaxiPath` neighbours.** A flight master connects to many nodes; a scripted quest flight connects to one, and boat and zeppelin nodes to none. Removes 658 of 1464 — every `Filming` and `Quest - …` row.
3. **`TaxiNodes.Flags` must say the node is shown to players.** The low two bits are the factions it appears for; bit `0x400` marks the client's own scripted and internal nodes. Every row the name-based rules had been chasing carries it — `Grand Rampart` is 1027, the Exile's Reach troll bat 1024 — and no real flight master does: Stormwind is 1, Orgrimmar 2, Maruukai 3. Removes 41.
4. **A containing zone box**, smallest wins.
5. **A `Place, Zone` name must corroborate that zone.** Zone boxes overlap at borders, which is how `New Kargath, Badlands` was filed under Searing Gorge, 1103 yards away. Removes 187 more. A name with no comma neither confirms nor contradicts, so it is kept but ranked below one that confirms — which is what makes Broken Shore pick `Vengeance Point, Broken Shore` over the bare `Dalaran` sitting 1332 yards away.

Getting here took two wrong turns, both caught by review. A clause rejecting a `Place, Zone` node when the stated zone was also represented destroyed **eight** of the ten capitals this branch now pins — Orgrimmar, Ironforge, Thunder Bluff, Darnassus, Shattrath, Dazar'alor, Valdrakken and Dornogal — because an entry claims the zone the *geometry* put it in, not the one its name mentions. (Earlier revisions of this description said six, while listing seven names. Both were wrong.) And a rule requiring bare names to *be* the zone dropped Dazar'alor, Undermine and Isle of Quel'Danas, whose flight masters are called `The Great Seal`, `The Incontinental Hotel` and `Shattered Sun Staging Area`. Both were replaced by the `Flags` test, which is a field the client sets rather than another guess at names.

The generator is committed as `scripts/generate_flight_points.py`, writes the Lua file directly, and reproduces the committed data byte for byte. It refuses to write below 100 zones, names its required columns, and rejects ragged rows — a truncated `TaxiPath.csv` used to produce an empty data file and exit 0.

## Measured, corrected

Warrior with no teleports, routing to Stormwind, across the **94 flight zones that have a graph node** rather than a nine-zone sample:

**18 improved, 63 unchanged, 0 worse**; the remaining thirteen have no route to Stormwind in either state. Silithus 855s → 542s, Mount Hyjal 705s → 418s (one of the seven zones that had been excluded), Orgrimmar 645s → 514s, Badlands 307s → 185s.

(An earlier revision quoted a Mechagon pair that a second measurement did not reproduce — 480 → 283 against 475 → 278. The improvement is the same either way; the absolute pair depends on where inside Stormwind the route ends, so it is dropped rather than restated.)

(An earlier revision of this description called map 81 Tanaris. It is Silithus; Tanaris is map 71, 492s → 451s.)

`771b9c6` changed every `x`/`y` in the data file — see the coordinate section above — without changing a single route time, because edge weights are computed from `worldX`/`worldY` and nothing read the zone coordinates. That is precisely why the defect survived four rounds: wrong data with no consumer. The waypoint fix gave them a consumer, which is what finally made a wrong value visible.

## Still not what this fixes

Stranded zones are unchanged: 26 of the zones in `QR.ZoneToContinent` have no route to Stormwind, with flight and without. Flights do not cross world maps, which is correct game behaviour, so Pandaria and Draenor still have no route out. That is #21, and it needs portal destinations that only a live client can supply. ([#21](https://github.com/CybotTM/wow-quickroute/issues/21) says 21 rather than 26; that count came from enumerating the 153 zones the addon knows rather than this table. The delta is what matters here, and it is zero either way.)

## Verified

**Eight review rounds**, each running its own defect injections against the previous round's tests. Every round found something real, and five of them found it in the previous round's *fix* or in a claim I had written:

| round | what it found |
|---|---|
| 1 | 5 defects; five injections left all 11671 assertions green |
| 2 | the fixes held; 8 more, incl. a portal clobber and hash-order anchors |
| 3 | my own fix had **deleted six capital cities**; four fixes had no test |
| 4 | **every `x`/`y` in the data was transposed and mirrored** since day one |
| 5 | the generator was outside CI entirely; the capitals still had no guard |
| 6 | the node ranking, the zone-box rules and a flight step's waypoint had no test |
| 7 | **two real defects** — the waypoint pointed at the wrong building, and seven zones with flight masters were excluded |
| 8 | **my round-7 waypoint fix was a regression** — half the edges pointed at the wrong zone |

Rounds 4 and 5 are the ones worth reading. The transposed coordinates survived because the test written to catch bad coordinates asserted `0 < x < 1` — and a swapped, mirrored unit square is still the unit square, so the predicate held identically for all 134 wrong entries.

What is guarded now: the addon-side flight code by named Lua tests, and the generator by `tests/test_generate_flight_points.py` plus a CI job, because `ci.yml` had no Python job at all. Dropping any of the five filters, either half of the ranking, the zone floor, un-mirroring an axis or transposing the projection each fails a named test — each verified by the injection it exists for, run in isolation. Ten capitals are pinned by uiMapID, and the coordinate anchors are two rather than one — Stormwind sits near the diagonal, where a transposition moves it 0.028 and slips through; Undercity does not — this repo surveys its bank to within 0.01 of the flight master.

Five guards cannot be reddened by any injection: the teleport guard in `WriteFlightEdge` and `nodeA ~= nodeB` are unreachable by construction and say so in their comments; the `master.x and master.y` check, the `0 < x < 1` coordinate filter and the `[…]` strip are inert on the current tables — they drop nothing today. None is counted as coverage. (An earlier revision of this description listed the faction half of the `Flags` test as a third. That was wrong — it has a test; what is true is only that it drops nothing on the current tables.) Two others turned out to be dead code and were removed instead of left standing as false coverage — a redundant sort, and a neutral-continent wildcard the preceding early return already handled.

13066 Lua assertions, 0 failures, in both file orders; 25 Python tests. Nine review rounds; the last one ran 44 injections and found no correctness defect. Luacheck 0 warnings / 0 errors in 72 files — run with a locally unpacked 1.1.2, control-tested against a file with known defects first; CI pins 1.2.0. The generator reproduces the committed data byte for byte.


CybotTM:
Merging this on the instruction "wiederhole review und fixes bis keine neuen Befunde mehr auftauchen, merge wenn alles grün / danach weiter mit den offenen Punkten", read together with the goal to fix the remaining issues. I am stating that reading explicitly rather than assuming it, because the merge is otherwise mine to queue and not to perform.

Copilot review is unavailable — the account is out of its monthly quota, which two attempts earlier in this session confirmed. What stands in for it is nine independent review rounds, each one running its own defect injections against the previous round's tests rather than reading the diff.

Every round found something real. Five found it in the previous round's fix or in a claim I had written:

| round | what it found |
|---|---|
| 1 | 5 defects; five injections left all 11671 assertions green |
| 2 | the fixes held; 8 more, including a portal clobber and hash-order anchors |
| 3 | my own fix had deleted eight capital cities; four fixes had no test |
| 4 | every `x`/`y` in the data had been transposed and mirrored since day one |
| 5 | the generator was outside CI entirely; the capitals still had no guard |
| 6 | the node ranking, the zone-box rules and a flight step's waypoint had no test |
| 7 | two real defects — the waypoint pointed at the wrong building, seven zones were excluded |
| 8 | **blocked the merge**: my round-7 waypoint fix was a regression, half the edges pointing at the wrong zone |
| 9 | not blocking; 44 injections, 33 red, no correctness defect |

Round 8's four blockers are cleared and round 9 verified each by execution rather than by reading: the waypoint is the departing zone's flight master on 599 of 599 edges, including both halves of every mirrored pair; both new tests redden their own defect and neither is redundant; the dropped `navMapID` assignment is genuinely dead; the boarded-transport clause is genuinely load-bearing; the removed edge fields have no readers.

Three claims round 9 found false have been corrected rather than argued with — a wrong reason in a code comment about the `Val` placeholder map, a "63%" figure that reproduced on no basis, and a citation to release notes that say no such thing.

What is deliberately not fixed here, filed instead: [#22](https://github.com/CybotTM/wow-quickroute/issues/22) (faction-specific flight points offered to both factions — pre-existing, and the fix needs a decision) and [#23](https://github.com/CybotTM/wow-quickroute/issues/23) (waypoint map and coordinates can come from different nodes — two of its three lines predate this branch).

Gates on `8b95863`: 4/4 checks green, 0 unresolved threads, `mergeStateStatus: CLEAN`. 13066 Lua assertions in both file orders, 25 Python tests, luacheck 0 warnings / 0 errors in 72 files, and the generator reproduces the committed data file byte for byte.

CybotTM:
Self-review: 8b95863e1df0f78ef4e2e143f4dfcaa7673188c6

Issue #23, opened 2026-08-31
Route step waypoints can take a map ID and coordinates from different nodes


CybotTM:
Three lines in `BuildSteps` decide where a route step's waypoint goes, and injecting a defect into any of them moves hundreds of waypoints onto the wrong map while all 13066 assertions stay green. Two are older than the flight work; one arrived with it.

## The three

**`PathCalculator.lua:1707`** — the flight-master block is guarded by `edge.edgeType == "flight"`. Drop the guard and it applies to every step type: 220 of 661 sampled steps get a different waypoint, and a walk step from the player to `Ratchet to Booty Bay (Start)` moves from (0.6330, 0.3810) to (0.5304, 0.4364) — Durotar's flight master, on a step that is not a flight. This one is new with #20 and its current value is correct; it simply has no test.

**`PathCalculator.lua:1615`** — `step.destMapID = edge.data.toMapID or …`. 1212 of 3029 edges that carry `data.toMapID` disagree with their to-node's own map: `Dungeon: Eye of Eternity → Dungeon: Azjol-Nerub` has a node on map 115 and `data.toMapID` 114. Removing the `or` fallback is green. Pre-existing.

**`PathCalculator.lua:1679`** — the default `navMapID = destMapID` for step types that are not boarded transports. Setting it to `fromMapID` instead changes 572 of 661 sampled steps and is green. Pre-existing.

## Why this is one issue and not three

The shape is the same in all three: a map ID and a coordinate pair are set from different sources, and nothing checks they belong together. The flight step is only guarded now because #20 added a test that compares the waypoint against a known position; no such check exists for portals, boats, zeppelins, the tram or plain walking.

## What closing it looks like

An invariant rather than three more spot tests: for every step of every type, `navX`/`navY` must lie in `navMapID`, and `navMapID` must be the map of the node the player is being sent to. The route fixtures to run it over already exist in `tests/test_pathfinding.lua`.

Treat that as the problem statement, not the design — whoever picks it up should check whether `destMapID`'s fallback is load-bearing for some step type before removing it.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_019wx8ueHuqUidp5yybDQ2CN)_

Issue #22, opened 2026-08-31
Faction-specific flight points are offered to both factions


CybotTM:
`QR.FlightPoints` holds one flight point per zone, and some of those points are reachable by only one faction. The addon offers them to both.

`TaxiNodes.Flags` carries the faction bits — 1 Alliance, 2 Horde, 3 both — and the generator uses them only to reject nodes shown to nobody (`Flags & 0x3 == 0`). The value is then discarded, so a zone whose chosen representative is faction-specific prices a flight that half the players cannot take.

Concrete, from the 141 entries shipped in #20:

| zone | node | faction |
|---|---|---|
| Northern Stranglethorn | `Grom'gol, Stranglethorn` | Horde only |
| The Cape of Stranglethorn | `Booty Bay, Stranglethorn` | Alliance only |
| Ruins of Gilneas | `Forsaken Forward Command, Gilneas` | Horde only |
| Stormwind City | `Stormwind, Elwynn` | Alliance only |
| Orgrimmar | `Orgrimmar, Durotar` | Horde only |

The capitals were already like this before #20 and nobody has reported it, which is a fair signal about severity: an Alliance player is not routing out of Orgrimmar anyway, and the graph reaches those zones by other edges. The three Stranglethorn and Gilneas rows are the ones that could bite, because those zones are ordinary questing zones a player of either faction passes through.

## What closing it looks like

The information is already in the input. `generate_flight_points.py` reads `Flags` at the filter step; carrying the faction bits into the emitted entry costs one field. `PathCalculator:AddFlightEdges` already knows the player's faction — `QR.PlayerInfo` resolves it — so the edge would simply not be written where the point is for the other side.

Two things to check rather than assume when picking this up:

- Whether a zone has a second, opposite-faction flight point that the one-entry-per-zone collapse is currently discarding. If it does, the fix is to keep both and choose at graph-build time rather than to drop the edge.
- Whether `Flags` really means "visible to this faction" for taxi nodes rather than something narrower. The bit meanings here were derived by measurement against known nodes (Stormwind 1, Orgrimmar 2, Maruukai 3), not from documentation.

## Verifying a fix

`tests/test_generate_flight_points.py` has fixtures for faction-flagged nodes already. A test should build the graph twice, once per faction, and assert that a one-sided zone appears in exactly one of them.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_019wx8ueHuqUidp5yybDQ2CN)_

Issue #21, opened 2026-08-31
Pandaria and Draenor have no modelled way back out (21 stranded zones)


CybotTM:
21 zones have no route to anywhere else, for a character with any set of teleports. Measured across all 153 zones the addon knows, routing to Stormwind: 29 zones had no route before #18, 21 after, and #20 leaves that number unchanged.

The 21 are Pandaria and Draenor. Both continents are reachable — `QR.PortalHubs` models the way in — and neither models a way back out. Flight paths do not help here: no flight-master taxi path leaves either continent. That is true of these two continents specifically, not of flights in general — see the correction below.

## Why this is not fixable from the client tables

The other data in this addon came from DB2 exports, which is why #1 and #2 could be built without a live client. This one cannot be, and the reason is specific rather than general.

A portal's destination is not in any client table at all. [WoWDBDefs](https://github.com/wowdev/WoWDBDefs) carries 1805 DB2 definitions, untruncated, and `SpellTargetPosition` is not among them — `SpellTargetRestrictions` is. It is a **server-side** table: [TrinityCore](https://trinitycore.atlassian.net/wiki/spaces/tc/pages/2130126/spell_target_position) keeps `spell_target_position` in its world database, populated from packet captures. `SpellEffect` gives the effect type but not the coordinates it lands on.

So no client-data exporter will ever carry this, and switching to a different one would not help. The destination map and position of each return portal has to be observed from a live client.

## What closes it

For each return portal in Pandaria and Draenor: the map it lands on, and where. Two ways to get it:

- `/qrverifymap`, which shipped in 1.12.0 — stand at the portal's destination, run it, and the first line is the map ID the client reports. One portal at a time, by hand.
- `/qrsurvey`'s doorway recorder — records both ends of every portal walked through, while playing. Nothing to run at each portal.

Treat that as the problem statement, not the design: whoever picks this up should re-derive how the hubs are modelled rather than assume the current `PortalHubs` shape is the right place for it.

## Verifying a fix

`tests/test_pathfinding.lua` has the stranding sweep. A fix should move the count below 21 and the test should say which zones it expects to be reachable, not just that the number went down.

## Corrections to this issue

Two statements above were wrong when this was filed and have been rewritten in place. Both were load-bearing, which is why they are named rather than quietly replaced.

[...]

PR #24, opened 2026-08-31
chore: release 1.13.0

CybotTM:
Cuts 1.13.0 from the flight-network work merged as [#20](https://github.com/CybotTM/wow-quickroute/pull/20). Two files: the TOC's `## Version` field, the addon's only version surface, and the `[Unreleased]` heading, which becomes `[1.13.0] - 2026-08-31`.

Minor rather than patch, because the flight network is an addition.

## What it carries

**Added** — flight paths are part of the travel graph, closing [#2](https://github.com/CybotTM/wow-quickroute/issues/2). 141 zones with a flight master, derived from the client's own tables rather than surveyed, connected only where the player has actually discovered them, and each flight step's waypoint is the flight master itself.

Measured for a character with no teleports routing to Stormwind, across the 94 flight zones that have a graph node: 18 get a faster route, 63 are unchanged, none get a slower one. Silithus 855s → 542s, Mount Hyjal 705s → 418s, Orgrimmar 645s → 514s, Badlands 307s → 185s.

## Still open, deliberately

- [#21](https://github.com/CybotTM/wow-quickroute/issues/21) — Pandaria and Draenor still have no modelled way out. Flights do not cross world maps, which is correct game behaviour, so this was never what the flight network would fix. It needs return-portal destinations that only a live client can supply.
- [#22](https://github.com/CybotTM/wow-quickroute/issues/22) — faction-specific flight points are offered to both factions. Pre-existing, and the fix needs a decision about whether to keep two points per zone.
- [#23](https://github.com/CybotTM/wow-quickroute/issues/23) — a route step's waypoint map and coordinates can come from different nodes. Two of its three cases predate this branch.
- [#5](https://github.com/CybotTM/wow-quickroute/issues/5) and [#3](https://github.com/CybotTM/wow-quickroute/issues/3) — both need one person standing in one place with `/qrverifymap`.

## Verified on this branch

- 13066 Lua assertions, 0 failures, in both file orders
- 25 Python tests for the data generator
- Luacheck: 0 warnings / 0 errors in 72 files
- `## Version: 1.13.0`, `## Interface: 120100`
- The generator reproduces `QuickRoute/Data/FlightPoints.lua` byte for byte

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+## [1.13.0] - 2026-08-31
+
 ### Added
 - Flight paths are part of the travel graph. 141 zones with a flight master, positioned from the client's own tables and connected where the player has discovered them. A flight step's waypoint is the flight master itself. Measured for a character with no teleports routing to Stormwind: 18 zones get a faster route and none get a slower one; Silithus drops from 855 to 542 seconds and Mount Hyjal from 705 to 418.
 
--- a/QuickRoute/QuickRoute.toc
+++ b/QuickRoute/QuickRoute.toc
@@ -2,7 +2,7 @@
 ## Title: QuickRoute
 ## Notes: Shows the shortest path to your destination using teleports and portals
 ## Author: Sebastian Mendel
-## Version: 1.12.0
+## Version: 1.13.0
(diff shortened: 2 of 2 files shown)


CybotTM:
Self-review: ba23eb855d317afbf714e6725772ae1688ea74ee

PR #25, opened 2026-08-31
feat: give each faction its own flight masters

### LINKED ISSUE #22: Faction-specific flight points are offered to both factions

CybotTM:
Closes [#22](https://github.com/CybotTM/wow-quickroute/issues/22).

## The issue understated it

It named five zones. Measured: **74 of 141** flight zones were represented by a node only one faction can walk up to, and the addon offered every one of them to both.

The number that decided the design:

| | zones |
|---|---|
| a flight master for **each** faction | 112 |
| Alliance only | 17 |
| Horde only | 12 |
| **the other faction's master discarded by the one-entry collapse** | **45** |

So a single entry with a `faction` field was not enough. It would have handed 45 zones to one faction and taken them from the other. The data carries both points now, the second as `alt`.

Some pairs are the same town — Booty Bay, Gadgetzan, Everlook and Moonglade each have two nodes a few yards apart. Others are nowhere near each other:

| zone | Alliance | Horde |
|---|---|---|
| The Hinterlands | Aerie Peak (north-west) | Revantusk Village (south-east) |
| Ashenvale | Astranaar | Splintertree Post |
| Arathi Highlands | Refuge Pointe | Hammerfall |
| Desolace | Nijel's Point | Shadowprey Village |

## One accessor knows the shape

`PathCalculator:FlightPointFor(uiMapID)` returns the point this player can use, or `nil` where the zone has no master for their faction — a real answer for 29 zones. A faction that is neither Alliance nor Horde — `"Neutral"` for an unchosen pandaren — gets the primary entry, which is what the addon did before this filter existed.

Note what that does **not** cover, because an earlier revision of this description claimed it did: a client that has not answered yet. `UnitFactionGroup` returns nil there and `GetFaction` turns nil into `"Alliance"`, so such a character is filtered as Alliance and never reaches the fallback. It is for a genuine third value, not for silence.

Both consumers go through it: `AddFlightEdges` prices edges from the player's own points, `BuildSteps` puts the waypoint on the master they can actually reach.

**Measured per faction:** Alliance 129 usable zones / 473 flight edges, Horde 124 / 417.

## Two review rounds, and both found the blocker in my own work

**Round 1 blocked the merge.** `UnitFactionGroup` returns `"Neutral"` for a pandaren who has not picked a side. `FlightPointFor` tested `not faction`, which can *never* be true — `GetFaction` falls back to `"Alliance"` — so a neutral character matched no branch at all and went from 141 usable zones and 479 flight edges down to **67 and 87**. Anything that is not one of the two factions now gets the primary entry, which is what the addon did before this filter existed.

That round also found `emit()` had **no coverage whatsoever**: deleting the entire alt tail — all 45 alternates, the whole payload of this change — or stamping `faction = "both"` on every row was green on both suites, because the Python tests only exercised `build()` and the Lua tests read a committed file no job regenerates. And the alternate carried the *primary's* `continentID` while a comment, a test and this description all called it self-contained.

**Round 3 found a fifth.** "As Horde it changes 17 zones" describes the zones with an Alliance primary and a Horde alternate — that set is **26**. 17 is the Alliance-only zones, a different set, where a Horde player gets nothing rather than something else. Corrected.

**Round 2 found a fourth wrong claim, in the commit that was fixing three.** "386 of 473 flight edges" mixed the Alliance baseline into a neutral measurement. A neutral character has **479** edges on main, not 473 — `AddZoneNodes` withholds the faction capitals from them — so the loss was **392 of 479**. Corrected in both comments that carried it.

It also found the `continentID` fix half-applied: `WorldMapMixesContinents`, the one index that consumes `continentID`, still read only primaries. Both halves feed it now.

## Verified

Every injection aimed at this change reddens a named test, run one at a time in its own tree:

| injection | caught by |
|---|---|
| ignore the faction / never return the alternate | the accessor tests |
| graph or waypoint reads the primary entry | pricing and Horde-waypoint tests |
| generator drops the alternate, or collapses both factions | the generator tests |
| `emit()` drops the alt tail, or stamps `"both"` everywhere | the five new emit tests |
| the alternate takes the primary's world map | a fixture with the two nodes on different world maps |
| the neutral regression, reintroduced | the neutral test |
| the generated header reverts to "one entry per zone" | the header test |

Two guards cannot be reddened and say so rather than being counted: `alt.faction == faction` (primary and alternate are always opposite) and the `QR.PlayerInfo` nil check. Feeding the alternate into the continent index is a third — no shipped zone has its two nodes on different world maps, so it changes nothing today.

13496 Lua assertions, 0 failures, in both file orders; 34 Python tests; luacheck 0 warnings / 0 errors in 72 files; the generator reproduces the committed data file byte for byte.

## Not addressed — filed as [#27](https://github.com/CybotTM/wow-quickroute/issues/27) and [#28](https://github.com/CybotTM/wow-quickroute/issues/28)

- **A pandaren choosing a side is an in-session faction change** ([#27](https://github.com/CybotTM/wow-quickroute/issues/27)) — and it is exactly the character the blocker above was about. `PlayerInfo:GetFaction` caches on first read and nothing invalidates it, so such a character keeps `"Neutral"` for the session. The outcome is no worse than main (they see everything, as they did before), but an earlier revision of this description claimed faction cannot change within a session, which is wrong for them. The hook is `PlayerInfo:GetFaction`, not `FlightPointFor`.
- **`AddZoneNodes` uses the opposite rule** ([#28](https://github.com/CybotTM/wow-quickroute/issues/28)) for a non-faction value: a neutral character gets only neutral capitals there, while `FlightPointFor` now gives them everything. Pre-existing and byte-identical to main, and arguably `AddZoneNodes` is the stricter, more correct one — but the two disagree, and its faction filter is itself uncovered.
- **Northern Stranglethorn has no Alliance flight master in the data** ([#28](https://github.com/CybotTM/wow-quickroute/issues/28)), so an Alliance player now gets no flight edge for map 50 where main gave them a Horde-priced one. The cause is upstream: Rebel Camp's world position falls inside the smaller Duskwood box, so the smallest-box rule files it under Duskwood and the name check drops it.


CybotTM:
Copilot review is unavailable — the account is out of its monthly quota, recorded earlier in this session. What stands in for it is four review rounds, each running its own defect injections against the previous round's tests rather than reading the diff. Every round found something real, and every one of them found it in my own work.

| round | what it found |
|---|---|
| 1 | **blocked** — a neutral pandaren lost 74 of 141 zones and 392 of 479 flight edges; `emit()` had no coverage at all; the alternate carried the primary's `continentID` |
| 2 | **blocked** — the commit fixing three wrong claims contained a fourth: "386 of 473" mixed the Alliance baseline into a neutral measurement; `WorldMapMixesContinents` still read only primaries |
| 3 | one wrong count: "changes 17 zones" describes a set of 26; 17 is the Alliance-only zones, a different set |
| 4 | **the merge resolution dropped two lines of this PR** — see below |

The last one is worth reading. Merging main in produced an interleaved conflict, because both branches had inserted tests at the same anchor. I rebuilt the file from main's version plus this branch's *new* blocks, and ran a title census afterwards: 81 titles, none lost, none invented. The census could not see the defect — it counts blocks, it does not compare them. Two *shared* blocks had been edited by this branch and were silently restored to their base form: the two accessor conversions, in the tests whose whole subject is the thing this PR exists to change.

Demonstrated rather than argued: give the Badlands two faction masters at different positions, leaving the world coordinates alone, and the merged version fails twice — both times **rejecting a correct waypoint**, because it compares against the primary entry while production correctly uses the player's own point. Restored, the same data is green. It does not fire today because both zones those tests touch are `faction = "both"`.

Gates on `f498516`: 4/4 checks green, 0 unresolved threads, `mergeStateStatus: CLEAN`. 13630 Lua assertions in both file orders, 34 Python tests, luacheck 0 warnings / 0 errors in 72 files, and the generator reproduces the committed data file byte for byte.

Deliberately not fixed here, filed instead: [#27](https://github.com/CybotTM/wow-quickroute/issues/27) (a pandaren who picks a side keeps the faction they logged in with) and [#28](https://github.com/CybotTM/wow-quickroute/issues/28) (overlapping zone boxes drop Rebel Camp, and `AddZoneNodes` uses the opposite rule for a non-faction character). Both need a decision rather than a patch.

CybotTM:
Self-review: f498516b59409192f950269996153a702a4cb77d

PR #26, opened 2026-08-31
test: pin where a route step's waypoint comes from

### LINKED ISSUE #23: Route step waypoints can take a map ID and coordinates from different nodes

CybotTM:
Closes [#23](https://github.com/CybotTM/wow-quickroute/issues/23).

## Measured before deciding, and it changed what to do

The issue asked whether `destMapID`'s fallback is load-bearing before removing it. Measured on the built graph:

- **No edge carries `toX` or `toY` at all.** The portal descriptor that has them is nested under `portalData`, so those two branches were reading a field nothing writes into `edge.data`.
- **No edge with a `toMapID` lacks a to-node**, so the fallback never fires either — the node always wins.

Both are inert on the real data. So the mixing hazard is **latent, not live**: this is a test change plus a dead-branch removal, not a behaviour fix. Nothing a player sees changes.

## What was removed, and what stayed

The `toX`/`toY` branches are gone. They are the shape that has already cost this addon four review rounds — a field nobody reads is a field nobody notices is wrong — and they could overwrite `destX`/`destY` unconditionally while `destMapID` stayed with the node. That is precisely how a step comes to name one node's map and another node's position.

The `toMapID` fallback stays, with a comment saying it does no work today and why it is kept: for an edge whose to-node is missing.

## Five tests, because no single shape reaches all of it

**The waypoint invariant** walks routes of six step types — walk, travel, portal, boat, tram, flight — and asserts each waypoint's map and position belong to one thing: the node travelled to, the node departed from, the flight master, or the destination the caller asked for. That last case is not a graph node, which is why the first version of this test failed on a perfectly legitimate final step.

**Teleport steps are out of scope there, deliberately**, and covered by two tests of their own instead. A teleport lands where its spell puts you, which is neither endpoint, and the edge carrying those coordinates is gone by the time a finished route can be inspected.

**The destination test could not reach its own subject by sampling.** **1212 of 3029** edges carrying a `toMapID` disagree with their to-node's map — 422 with a dungeon node on both sides, 773 on one, 17 on neither — and an ordinary route to a capital touches none of them. Two rounds of widening the destination list still missed them. So it drives `BuildSteps` directly, once per disagreeing edge, which is the only way to make the rule visible: the node has to win.

## Two review rounds, both finding the problem in my own work

**Round 1 blocked the merge.** The invariant offered the caller's requested destination as a valid source for *every* step. So moving every waypoint onto the route's final destination — a player standing in Ironforge pointed at a spot in Orgrimmar, on another continent — passed with all 13073 assertions green in both file orders. That is the most obvious waypoint defect there is, and the test written to catch it waved it through. `requested` is legitimate for the last step only, whose to-node `CalculatePath` removes before returning; it is scoped there now.

That round also caught three claims of mine: "walks routes of every step type" is six of eight, the destination test asserted only the map despite its name, and "the 1212 disagreeing edges are between dungeon nodes" is true of 422 of them, one-sided for 773, and false for 17.

**Round 2 refuted a conclusion I had written down as fact.** I documented the three teleport overrides as untestable, having failed to build a route that reaches them. The first half was right — a routed teleport cannot reach them — but the conclusion did not follow: the sibling test two hundred lines up drives `BuildSteps` directly for exactly that reason, and **its fixture already had the node and the landing disagreeing**. It just asserted neither. Three assertions closed it. A second test, with the node on map 84 and the landing on 125, closes the map override too.

It also found the data test missing `DungeonTeleportSpells`, which `PlayerInventory` scans and hands to `BuildSteps` like the other four. Checked entries: 108 → 210.

## Verified

Every injection I can construct against `BuildSteps` reddens a named test, each run alone in its own tree:

| injection | result |
|---|---|
| edge data overrides the node for `destMapID` | red |
| default `navMapID` takes the departure map | red |
| flight guard dropped; flight-master position dropped | red |
| `navX`/`navY` from the wrong node, or transposed | red |
| boarded transports lose their from-node | red |
| **every waypoint moved onto the final destination** (the round-1 blocker) | red |
| **only the first waypoint moved** | red |
| **all three teleport overrides deleted**, or the coordinates alone, or the map alone | red |
| a teleport entry with a landing map but no coordinates | red |

13200 Lua assertions, 0 failures, in both file orders; 25 Python tests; luacheck 0 warnings / 0 errors in 72 files.

### DIFF
--- a/QuickRoute/Core/PathCalculator.lua
+++ b/QuickRoute/Core/PathCalculator.lua
@@ -1609,17 +1609,19 @@ function PathCalculator:BuildSteps(path, edges)
             step.destY = toNodeData.y or 0.5
         end
 
-        -- Also get from edge data if available
-        if edge.data then
-            if edge.data.toMapID then
-                step.destMapID = step.destMapID or edge.data.toMapID
-            end
-            if edge.data.toX then
-                step.destX = edge.data.toX
-            end
-            if edge.data.toY then
-                step.destY = edge.data.toY
-            end
+        -- The edge only fills a gap, never overrides. It used to be allowed to
+        -- overwrite destX and destY unconditionally while destMapID stayed
+        -- with the node, so a step could name one node's map and another
+        -- node's position -- the shape that put a flight waypoint in the wrong
+        -- zone twice. Nothing writes toX or toY into edge.data (the portal
+        -- descriptor that has them is nested under portalData), so those two
+        -- branches were reading a field that is never set; they are gone.
+        --
+        -- The toMapID fallback stays and currently never fires: every edge
+        -- carrying it has a to-node, so the node's map always wins. It is here
+        -- for an edge whose node is missing, not because it does work today.
+        if edge.data and edge.data.toMapID then
+            step.destMapID = step.destMapID or edge.data.toMapID
         end
 
         -- Get localized display names for source and destination nodes
--- a/tests/test_data_validation.lua
+++ b/tests/test_data_validation.lua
@@ -1535,3 +1535,54 @@ T:run("Data: FlightPoints zones are zones, not continents", function(t)
             "map " .. tostring(uiMapID) .. " is a zone")
     end
 end)
+
+-------------------------------------------------------------------------------
+-- Teleport landing coordinates
+-------------------------------------------------------------------------------
+
+T:run("Data: teleport landings are all-or-nothing", function(t)
+    -- BuildSteps applies a teleport's mapID, x and y under three separate
+    -- guards, each overriding the node. An entry with a mapID but no x or y
+    -- therefore produces a step naming one place's map and another place's
+    -- position -- the exact mix the waypoint invariant exists to prevent, and
+    -- the one case that invariant cannot see, because the edge carrying the
+    -- landing data is gone by the time a finished route can be inspected.
+    --
+    -- So it is checked here instead, at the source: a teleport that says where
+    -- it lands has to say it completely.
+    --
+    -- This covers the DATA, not the code that applies it. Deleting the three
+    -- overrides in BuildSteps is still invisible to the suite -- reaching them
+    -- needs a route whose teleport lands somewhere other than the destination
+    -- the caller asked for, and no fixture here produces one. Written down
+    -- rather than papered over with a test that would not fail.
+    local tables = {
+        TeleportItemsData = QR.TeleportItemsData,
+        ClassTeleportSpells = QR.ClassTeleportSpells,
+        RacialTeleportSpells = QR.RacialTeleportSpells,
+        GeneralTeleportSpells = QR.GeneralTeleportSpells,
+    }
+    local checked, withMap = 0, 0
+    for label, data in pairs(tables) do
+        for id, entry in pairs(data or {}) do
+            if type(entry) == "table" then
+                checked = checked + 1
+                if entry.mapID then
+                    withMap = withMap + 1
+                    t:assert(type(entry.x) == "number" and type(entry.y) == "number",
+                        label .. "[" .. tostring(id) .. "] states a landing map ("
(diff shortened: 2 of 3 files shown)


CybotTM:
Copilot review is unavailable — the account is out of its monthly quota, recorded earlier in this session. What stands in for it is three review rounds, each running its own defect injections against the previous round's tests rather than reading the diff.

Two of the three found the problem in my own work:

**Round 1 blocked the merge.** The invariant offered the caller's requested destination as a valid source for *every* step, so moving every waypoint onto the route's final destination — a player standing in Ironforge pointed at a spot in Orgrimmar — passed with all 13073 assertions green in both file orders. The test written to catch that defect waved it through. It also caught three claims of mine: "every step type" is six of eight, the destination test asserted only the map despite its name, and the "between dungeon nodes" phrasing is true of 422 of 1212 edges.

**Round 2 refuted a conclusion I had written down as fact.** I documented the three teleport overrides as untestable, having failed to build a route that reaches them. The premise was right — a routed teleport cannot reach them, the edge is gone by then — but the conclusion did not follow: the sibling test two hundred lines up drives `BuildSteps` directly for exactly that reason, and its fixture already had the node and the landing disagreeing. It asserted neither. Three assertions closed it; a second test with the node on map 84 and the landing on 125 closed the map override too. That round also found the data test missing `DungeonTeleportSpells` (108 → 210 checked entries).

**Round 3 found nothing new** after 20 isolated mutations, four historical control runs, and a check of every number in the body and all three commit messages against the artifacts. It verified the round-1 blocker is green at `5131248` and red now, that all five round-1 injections still redden, and that the four greens among eleven fresh mutations are each either declared inert with a measurement behind it or outside this PR's touched surface.

Gates on `0e9bd33`: 4/4 checks green, 0 unresolved threads, `mergeStateStatus: CLEAN`. 13200 Lua assertions in both file orders, 25 Python tests, luacheck 0 warnings / 0 errors in 72 files.

CybotTM:
Self-review: 0e9bd33451d4700c7b08d1bc71ceb9752bcaaa24

Issue #28, opened 2026-08-31
Overlapping zone boxes drop a flight master, and two faction rules disagree


CybotTM:
Two things the flight-network work surfaced but did not fix, both about which zone a flight master is filed under.

## Northern Stranglethorn has no Alliance flight master in the data

`QuickRoute/Data/FlightPoints.lua` ships uiMap 50 as Horde-only (`Grom'gol, Stranglethorn`). The Alliance master for that zone is `Rebel Camp, Stranglethorn` — TaxiNodes ID 195, Flags 5 — and it is dropped by the generator.

Traced: its world position (−11344, −216.83) falls inside the Duskwood assignment box (area 4 860 087) as well as the Northern Stranglethorn one (area 11 206 529). The smallest-containing-box rule therefore files it under uiMap 47 (Duskwood), and the name-vs-geometry filter then throws it away, correctly, because `Stranglethorn` is not `Duskwood`.

The filter is doing its job. The geometry is what is wrong: two zone boxes overlap and the smaller one wins a point that does not belong to it.

Effect after #22: an Alliance player gets no flight edge at all for map 50, where before they got a Horde-priced one. Neither is right; the second is quieter.

## `AddZoneNodes` and `FlightPointFor` disagree about a non-faction value

`PathCalculator.lua:352-356` keeps a node when `nodeData.faction == "both" or nodeData.faction == playerFaction`. A neutral pandaren matches neither for a faction capital, so they get no Stormwind and no Orgrimmar node.

`FlightPointFor`, added in #22, does the opposite for the same character: not being one of the two factions means no filtering, so they get every zone's primary flight point.

Both are defensible on their own and they contradict each other. `AddZoneNodes` is arguably the correct one — a neutral pandaren cannot use a Horde flight master either — but that would mean tightening `FlightPointFor` rather than loosening `AddZoneNodes`, and tightening it is what caused the regression #22 had to fix. It needs a decision, not a patch.

`AddZoneNodes`' faction filter is also uncovered: replacing the condition with `true` leaves the whole suite green.

## What closing this looks like

For the first half: a rule that prefers the box whose zone the node's own name states, when the name states one, over the strictly smallest box. That is the same corroboration signal the generator already uses to reject rows — using it to *choose* between overlapping boxes rather than only to veto would fix Rebel Camp without weakening the New Kargath case. Worth measuring how many other zones change before adopting it.

For the second: pick one rule for a non-faction character and apply it in both places, with a test for each.

[...]

PR #30, opened 2026-08-31
fix: file a flight master under the zone its name states

CybotTM:
Part of [#28](https://github.com/CybotTM/wow-quickroute/issues/28). The zone-box half; the rest stays open — see below.

## Zone boxes overlap, and the smallest one is not always right

Rebel Camp sits inside Duskwood's assignment box as well as Northern Stranglethorn's, and Duskwood's is smaller. The smallest-wins rule therefore filed it under Duskwood, and the name check then correctly threw it away.

Of the 756 nodes that survive the earlier filters, **662 sit inside more than one box**, and the rule moves **185** of them somewhere other than the smallest.

(An earlier revision of this description said 197. That is the count for the *stated-zone-only* variant below — the one this change rejects — not for the rule as written. Sixth wrong figure in this work, and the sixth of the same shape: a number correct for a neighbouring set, used for the set the sentence describes.)

## Both halves of the name, because the obvious version is wrong

The rule now prefers a box the node's name corroborates — and it needs *both* halves of `Place, Zone`. Measured before adopting:

| rule | what it does |
|---|---|
| stated zone only | moves Stormwind's master to **Elwynn Forest**, Ironforge's to **Dun Morogh**, Orgrimmar's to **Durotar** — those names state the zone *around* the city |
| place part only | leaves `Darkshire, Duskwood` in **Deadwind Pass** |
| **both** | 431 nodes keep the smallest box because it is itself corroborated, **185 move** |

Examples of the 185: `Darkshire, Duskwood` from Deadwind Pass to Duskwood, `New Kargath, Badlands` from Searing Gorge to Badlands, `Ratchet, Northern Barrens` from Valley of Trials to Northern Barrens, `Garadar, Nagrand` from Zangarmarsh to Nagrand.

## And a second thing it exposed

Once New Kargath was correctly filed in the Badlands, the zone **split** into a Horde primary and an Alliance alternate — purely because New Kargath is better connected than Fuselight, which serves everyone. Both sides can walk up to Fuselight; the split cost 19 seconds on the Alliance route and gained nothing.

Ranking now puts "a node both factions can use" ahead of degree. Measured both orders across every zone-to-Stormwind route, for both factions:

| ranking | result |
|---|---|
| both-factions **ahead of** degree | 15 routes faster, 7 slower — **net 304s saved** |
| both-factions **behind** degree | 6 faster, 4 slower — net 149s |

Degree is a proxy for centrality, not for usefulness to the player.

## Against main

- **141 → 147 zones**, none removed
- **no zone loses a faction's access**; nine gain one
- 12 zone-faction pairs become routable that were not
- 15 routes faster, 7 slower, 0 lost

## What this does not fix

**Rebel Camp still does not give Northern Stranglethorn an Alliance flight master** — the case [#28](https://github.com/CybotTM/wow-quickroute/issues/28) opens with. It now lands on uiMap 224, the pre-Cataclysm "Stranglethorn Vale" map, because that is what its name says. 224 has no graph node, so the entry is inert rather than harmful, and map 50 stays Horde-only. That half of the issue, and the `AddZoneNodes` disagreement, stay open.

## Verified

Five defect injections, each on its own copy: smallest-always, either half of the name rule alone, dropping the both-factions rank, and inverting it. All five redden a named test.

13560 Lua assertions, 0 failures, in both file orders; 37 Python tests; luacheck 0 warnings / 0 errors in 72 files.


CybotTM:
Copilot review is unavailable — the account is out of its monthly quota, recorded earlier in this session. The independent review agent covering this PR ran for an hour without returning (previous rounds took 12–21 minutes) and was stopped. It was mid-finding when it stopped, and that finding was real: **"No quantity equals 197."**

It was right. 197 is the count for the *stated-zone-only* variant this change rejects; the rule as written moves **185**. Re-derived: 756 nodes survive the earlier filters, 662 sit inside more than one box, 197 under stated-only, 185 under the rule. Corrected in the generator comment and in this description. Sixth wrong figure in this work, and the sixth of the same shape — a number correct for a neighbouring set, used for the set the sentence describes.

So this is a documented self-review, and it says what it covers.

**Verified by injection**, each mutation on its own copy:

| injection | result |
|---|---|
| box selection back to smallest-always | red |
| name rule uses only the stated zone | red |
| name rule uses only the place part | red |
| drop the both-factions rank | red |
| invert the both-factions rank | red |

**Verified by measurement**, re-derived here rather than taken from the agent:

- 756 surviving nodes, 662 in more than one box, 185 moved
- 141 → 147 zones, none removed; no zone loses a faction's access, nine gain one
- 15 routes faster, 7 slower, 0 lost, across every zone-to-Stormwind route for both factions
- both ranking orders compared: both-factions ahead of degree saves 304s net, behind it 149s
- all six newly covered zones checked against `TaxiNodes.csv` — real flight masters, positions match, none carries the internal flag

**The merge with main was checked before committing, not after.** The three files [#31](https://github.com/CybotTM/wow-quickroute/pull/31) touched are byte-identical to main here; the three this branch owns carry its work; both sides' new tests are present by name. That comparison is what the previous merge in this repo skipped, and skipping it cost two lines of the branch being merged.

Gates on `a582477`: 4/4 checks green, 0 unresolved threads, `mergeStateStatus: CLEAN`. 13569 Lua assertions in both file orders, 37 Python tests, luacheck 0 warnings / 0 errors in 72 files, generator reproduces the data byte for byte.

Still open in [#28](https://github.com/CybotTM/wow-quickroute/issues/28) and stated in the description: Rebel Camp gives Northern Stranglethorn no Alliance master — it lands on the pre-Cataclysm map 224, which has no graph node, so the entry is inert rather than harmful.

CybotTM:
Self-review: a58247775e3bac709adb8c8d1be9da0985c0ce9a

PR #31, opened 2026-08-31
fix: a pandaren who picks a side is routed as that side

### LINKED ISSUE #27: A pandaren who picks a side keeps the faction they logged in with
`PlayerInfo:GetFaction` caches on first read and nothing ever clears it. `InvalidateCache` exists at `QuickRoute/Utils/PlayerInfo.lua:85` but no shipped code calls it and no faction-change event is registered anywhere in `QuickRoute/`.

For most characters that is fine — faction really is fixed for the life of the account. It is not fixed for a pandaren, who begins neutral and picks a side at the end of the starting experience. That is an in-session change, and it is exactly the character #22 turned out to be about.

## What goes wrong

After choosing Horde, the cached value stays `"Neutral"` until the next reload. With #22 merged, a neutral character is deliberately not filtered — they get every zone's primary flight point. So the newly-Horde player is offered Aerie Peak as The Hinterlands' flight master, and Astranaar for Ashenvale, both Alliance, for the rest of the session.

The outcome is no worse than it was before #22 (they saw everything then too), which is why this is a follow-up and not a blocker on that PR. But #22's description originally claimed "faction does not change within one session, so there is no invalidation to add", and that is wrong for this character.

## Where the hook belongs

Not in `PathCalculator:FlightPointFor`, which was the guess in #22's first draft. `PlayerInfo:GetFaction` is the single reader of the cached value, and `PlayerInfo:InvalidateCache` already exists to clear it — what is missing is anything calling it.

Two things to check rather than assume:

- Which event actually fires on the pandaren faction choice. `NEUTRAL_FACTION_SELECT_RESULT` is the plausible candidate but has not been verified against a live client, and the addon registers no faction events today.
- Whether the graph needs rebuilding as well as the cache clearing. `AddZoneNodes` and `AddFlightEdges` both read the faction at build time, so a stale graph would survive a cleared cache.

## Verifying a fix

`tests/test_pathfinding.lua` already switches `MockWoW.config.playerFaction` and calls `InvalidateCache` by hand — a test would fire the event instead and assert the graph changed, with `FlightPointFor(26)` moving from Aerie Peak to Revantusk Village.

CybotTM:
Closes [#27](https://github.com/CybotTM/wow-quickroute/issues/27).

## The one character whose faction changes

`GetFaction` cached its first answer and nothing ever cleared it — the addon registers no faction events, and no shipped code called `InvalidateCache`. For every character that is correct: faction is fixed for the life of the account.

Except for a pandaren, who begins neutral and chooses at the end of the starting experience. They kept `"Neutral"` for the rest of the session, and every faction-gated decision was made for a character who no longer existed.

## No event needed, which is the point

The issue suggested hooking a faction-change event, and flagged that the event name could not be verified offline. It does not need one. **`"Neutral"` is the only value that can change**, so it is the only one not cached. Alliance and Horde are cached exactly as before.

A test pins that half too: changing the underlying value *without* clearing the cache must **not** be picked up, because a chosen faction cannot change. Dropping the cache entirely fails it.

## The graph needed the same treatment

`AddZoneNodes` and `AddFlightEdges` both read the faction at build time, so a graph built before the choice describes the wrong character even once the cache is right. `CalculatePath` now compares the faction the graph was built for against the current one and rebuilds on a mismatch. Nothing else can change faction mid-session, so the comparison costs nothing the rest of the time.

## This also settles half of [#28](https://github.com/CybotTM/wow-quickroute/issues/28)

`AddZoneNodes` and `FlightPointFor` disagree about a character of neither faction — one withholds the faction capitals, the other filters nothing. With the faction updating the moment it is chosen, a neutral character only exists on the starting island, where neither rule reaches anything. The disagreement stays in the code and stops being observable.

## Verified

| injection | result |
|---|---|
| pin `"Neutral"` again (the original bug) | red |
| drop the cache entirely, so every read hits the API | red |
| do not rebuild the graph on a faction change | red |
| never record which faction the graph was built for | red |

13639 Lua assertions, 0 failures, in both file orders; 34 Python tests; luacheck 0 warnings / 0 errors in 72 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- A pandaren who picks a faction is routed as that faction straight away. The addon read the faction once and kept the answer, so a character who chose Horde was routed as neutral until the next reload.
+
 ## [1.14.0] - 2026-08-31
 
 ### Fixed
--- a/QuickRoute/Core/PathCalculator.lua
+++ b/QuickRoute/Core/PathCalculator.lua
@@ -274,6 +274,10 @@ function PathCalculator:BuildGraph()
 
     -- Only mark clean if all steps succeeded
     self.graphDirty = not buildSuccess
+    -- Remembered so CalculatePath can notice a faction change. Recorded even
+    -- on a failed build: a half-built graph belongs to the faction it was
+    -- attempted for.
+    self.graphFaction = QR.PlayerInfo and QR.PlayerInfo:GetFaction()
 
     -- Log graph stats
     local nodeCount, edgeCount = 0, 0
@@ -607,7 +611,17 @@ end
 -- @param destY number The destination Y coordinate (0-1)
 -- @return table|nil {path, totalTime, edges, steps} or nil if no path found
 function PathCalculator:CalculatePath(destMapID, destX, destY, destTitle)
-    -- Rebuild graph if needed
+    -- Rebuild graph if needed. Faction is part of "needed": AddZoneNodes and
+    -- AddFlightEdges both read it at build time, so a graph built before a
+    -- pandaren picked a side describes the wrong character. Nothing else can
+    -- change faction mid-session, so this comparison is free the rest of the
+    -- time.
+    local faction = QR.PlayerInfo and QR.PlayerInfo:GetFaction()
+    if self.graph and self.graphFaction and faction and self.graphFaction ~= faction then
+        QR:Debug("PathCalculator: faction changed from " .. tostring(self.graphFaction)
+            .. " to " .. tostring(faction) .. ", rebuilding the graph")
+        self.graphDirty = true
+    end
     if self.graphDirty or not self.graph then
         self:BuildGraph()
     end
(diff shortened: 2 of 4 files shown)


CybotTM:
Copilot review is unavailable — the account is out of its monthly quota, recorded earlier in this session. The independent review agent for this PR ran for an hour without returning (previous rounds took 12–21 minutes) and was stopped; its last line was a finding on the sibling PR, which I verified and corrected. So this is a documented self-review, and it says what it does and does not cover.

**Verified by injection**, each mutation on its own copy, in both file orders:

| injection | result |
|---|---|
| pin `"Neutral"` again (the original bug) | red — `Choosing a side mid-session takes effect without a reload` |
| drop the cache entirely, so every read hits the API | red — `A settled faction is still cached` |
| do not rebuild the graph on a faction change | red |
| never record which faction the graph was built for | red |

The second row is the one worth naming: the fix must not turn every faction read into an API call, so a test asserts that changing the underlying value *without* clearing the cache is **not** picked up for Alliance or Horde. Only `"Neutral"` is re-read.

**What this does not cover.** The event that fires when a pandaren chooses a side is not verified against a live client — deliberately, because the fix needs no event. If `UnitFactionGroup` were to keep returning `"Neutral"` after the choice, this would not help; that is the one assumption, and it is the same assumption every other faction-gated feature in the addon already makes.

Gates on `b70c4e8`: 4/4 checks green, 0 unresolved threads, `mergeStateStatus: CLEAN`. 13639 Lua assertions in both file orders, 34 Python tests, luacheck 0 warnings / 0 errors in 72 files.

CybotTM:
Self-review: b70c4e88c014b724bfee3ae75e8618ec3f265f34

Issue #33, opened 2026-08-31
Flight networks are guessed from geography; TaxiPath has them directly


CybotTM:
`SameFlightNetwork` decides whether two zones may have a flight edge by comparing their world map and the addon's own continent. Both are proxies for the thing that actually decides it: whether the client's taxi graph connects their flight masters. `TaxiPath` has that directly, and the connected components of the flight-master subgraph *are* the flight networks.

Filing this because it was built, measured, and **not shipped** — the measurement is the useful part, and it should not have to be redone.

## The proxy is wrong in both directions

**Not sufficient.** World map 530 holds Outland and the Burning Crusade starting zones. Without a second test the graph offered 50 flights nobody can take — this is why the addon-continent check exists.

**Not necessary.** 30 flight-master-to-flight-master taxi paths join two different world maps:

| world maps | paths | example |
|---|---|---|
| 0 ↔ 1190 | 10 | Shattered Beachhead, Blasted Lands ↔ Darkshire, Duskwood |
| 0 ↔ 530 | 9 | Light's Hope Chapel ↔ Tranquillien, Ghostlands |
| 1 ↔ 530 | 4 | The Exodar ↔ Rut'theran Village |
| 1795 ↔ 870 | 4 | Mistfall Village ↔ Shrine of Two Moons |
| 2552 ↔ 2601 | 2 | Dornogal, Isle of Dorn ↔ Gundargaz, The Ringing Deeps |
| 0 ↔ 1 | 1 | Sandy Beach, Vashj'ir ↔ Orgrimmar, Durotar |

So "flights do not cross world maps" — which I wrote in [#21](https://github.com/CybotTM/wow-quickroute/issues/21) and in two release notes — is false. It happens to be true of Pandaria and Draenor, which is why the stranding conclusion there still holds, but it is not a general rule.

## What the continent test is really for

That last row matters. A single Vashj'ir-to-Orgrimmar flight joins the Eastern Kingdoms and Kalimdor networks, so by connectivity alone Darnassus and the Sunken Temple are one edge apart. Replacing the continent test with the network test produced **468 cross-continent flight edges**.

They are not wrong about *reachability* — the game would auto-route that, at length. They are wrong about *price*: the weight here is a straight line divided by a speed, which is fair for a hop inside one landmass and nonsense across two. The continent test is doing pricing, not reachability, and its comment should say so.

## Why it was not shipped

[...]

PR #38, opened 2026-08-31
fix: file flight masters on the map the player is actually on

CybotTM:
Found while re-deriving the first half of [#28](https://github.com/CybotTM/wow-quickroute/issues/28), which turned out to be about something else. This is the defect that measurement uncovered instead.

## What it was

`UiMap` ships a second set of zone maps under `System = 2` — the same zones under different IDs, on assignment boxes covering the same world coordinates. They compete for every taxi node, and where one won the flight master was filed under an ID no player is ever on.

Seven of the 147 shipped flight points were:

| shipped | should be | zone | known to `ZoneAdjacency.lua`? |
|---|---|---|---|
| 1187 | 630 | Azsuna | shipped ID: no · correct ID: yes |
| 1569 | 1533 | Bastion | no · no |
| 2271 | 2248 | Isle of Dorn | no · yes |
| 2272 | 2214 | The Ringing Deeps | no · yes |
| 2273 | 2215 | Hallowfall | no · yes |
| 2480 | 2413 | Harandar | no · yes |
| 2270 | 2255 | Azj-Kahet | no · yes, and 2255 already had its own entry |

**Five of those IDs are ones the addon's zone tables do not know at all.** Those zones had a flight master in the data and no flight edge in the graph. It also accounts for the observation recorded in [#33](https://github.com/CybotTM/wow-quickroute/issues/33) that The Ringing Deeps has no node for a flight to attach to.

The coordinates move for three of them as well, because the projection now runs against the right zone's box.

## Measured, including the part that is not a win

Against `main`, both factions, every flight point discovered:

- flight edges: **509 → 527**
- zone-to-capital routes changed: **0 of 306**
- routes into the six affected zones changed: **0 of 1824**

So the data was wrong and the routing effect today is nil — the new edges never beat a portal. Stated rather than buried: this is a correctness fix, not a speed one. It is worth shipping because an entry keyed on a map the player is never on cannot become right later, and unlike [#33](https://github.com/CybotTM/wow-quickroute/issues/33) it costs no new data field and no runtime condition.

Entry count drops 147 → 146: the duplicate Azj-Kahet entry goes, and the correct one was already there.

## Verified

| injection | result |
|---|---|
| the `System` filter removed | the generator test names the duplicate; the shipped-data test lists all seven |
| `System` dropped from the required columns | the missing-column test — an export without it would otherwise yield no zones silently |
| regenerate without the filter, then run the Lua suite | both data guards red, all seven and all six named |

`System` is now a required column on `UiMap.csv` for that reason. The committed `FlightPoints.lua` is byte-identical to a fresh run of the committed generator.

13603 Lua assertions, 0 failures, in both file orders; 39 Python tests; luacheck 0 warnings / 0 errors in 72 files.

## Leaves issue 28 open

Its first half is about Rebel Camp, a different defect — and one the issue describes incorrectly. A comment there has the measurement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## [Unreleased]
 
 ### Fixed
+- Six zones had a flight master in the data and no flight edge in the graph. The client ships a second set of zone maps that repeats zones under different IDs, and the flight-point generator was filing masters under those: Azsuna as 1187 rather than 630, Isle of Dorn as 2271 rather than 2248, and the same for Bastion, The Ringing Deeps, Hallowfall and Harandar. Five of those IDs are ones the addon's zone tables do not know at all. Azj-Kahet had two entries and keeps the right one. No route changes as a result — 0 of 1824 measured routes into those zones and 0 of 306 to the faction capitals — but the graph gains 18 flight edges that were previously attached to nothing.
+
+### Fixed
 - Teleport items you are wearing work again. Eleven of them sit in slots the inventory scan never looked at — the six guild cloaks and Mountebank's Colorful Cloak on the back, three pairs of slippers on the feet, and the Blessed Medallion of Karabor on the neck — so while worn they counted as not owned and the panel drew a plain icon with nothing behind it. The slot list is now taken from the item data instead of being kept by hand.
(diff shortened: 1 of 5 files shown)


CybotTM:
Self-review: 35f8c3c22a2b460751b4e0af05e121adf6954a9e

PR #45, opened 2026-09-01
fix: add the General section header to the settings layout

### LINKED ISSUE #44: The "General" settings section header is created and discarded
`SettingsPanel.lua:105` builds the "General" section header and throws it away.

```lua
Settings.CreateElementInitializer("SettingsListSectionHeaderTemplate", { name = L["SETTINGS_GENERAL"] })
```

The return value goes nowhere. Every other control in that file is created through a function that takes the **category** and therefore registers itself — `Settings.CreateCheckbox(category, …)`, `Settings.CreateSlider(category, …)`, `Settings.CreateDropdown(category, …)`. `Settings.CreateElementInitializer(template, data)` does not take one, so it cannot add itself to anything; the caller has to.

## What supports it

- It is the **only** use of `CreateElementInitializer` in the addon, and the only use of `SETTINGS_GENERAL` outside `Localization.lua`.
- `SETTINGS_GENERAL` is translated in **all ten** locales — "General", "Allgemein", "Général", … Ten translations for a string that reaches nothing.
- No test covers it, and none could: `tests/mock_wow_api.lua:1662` stubs `CreateElementInitializer` as `function(template, data) return {} end`, so discarding the result looks identical to using it.

## What would settle it

One look at the QuickRoute page in the game's options: is there an "Allgemein" heading above the first checkbox, or do the checkboxes start straight away? That is the whole question — the code reads as a discard, but whether Blizzard's layout picks up a loose initializer some other way is a claim about their internals that cannot be checked from here.

## Why it is worth fixing beyond the missing heading

The same mechanism is how a custom element gets into a vertical-layout settings category — which is what the C2 options header from the design canvas needs. Getting this one right first means the header is built on a call that is known to work rather than on the same uncertainty.

CybotTM:
Closes [#44](https://github.com/CybotTM/wow-quickroute/issues/44).

## The header was built and thrown away

`Settings.RegisterVerticalLayoutCategory` returns the category **and its layout**, and the layout is what anything other than a checkbox, slider or dropdown has to be added to. Those three take the category and register themselves; a section header does not.

The code captured only the first return value, then called `Settings.CreateElementInitializer` and did nothing with the result. So the header never reached a layout, and ten translations of `SETTINGS_GENERAL` had nowhere to appear.

## Verified against the real client, not inferred

The issue was filed on reasoning: `CreateCheckbox(category, …)` takes a category, `CreateElementInitializer(template, data)` does not, so it cannot register itself. That is an inference about Blizzard's internals, and the issue said so.

It is now measured. Running the real Blizzard UI under [wow-ui-sim](https://github.com/Osso/wow-ui-sim):

```
layout        : table: 0x0000a949
AddInitializer: function: 0x0000604c
helper global : function: 0x00005d92
```

Two return values, `layout:AddInitializer` exists, and `CreateSettingsListSectionHeaderInitializer` is a plain global — which is the shape the Warcraft Wiki's "Creating a settings menu" documents and the code did not use.

## The mock encoded the same mistake

`tests/mock_wow_api.lua` had `RegisterVerticalLayoutCategory` return one value, so a discarded initializer looked exactly like a used one and **no test could have caught this**. It returns a layout now, recording the initializers it is given, and the section-header helper exists as the global it is.

## Verified

| injection | what reddens |
|---|---|
| the initializer built and discarded again | the section-header test — nothing reaches the layout |
| the layout not captured from the second return value | the same test, on the capture assertion |

## After the rebase onto main

[#48](https://github.com/CybotTM/wow-quickroute/pull/48) had meanwhile put the header element into the same function through `SettingsPanel:GetLayout(category)`; the rebase left two `layout` locals and the section header went to the second one. Both now go to the layout `RegisterVerticalLayoutCategory` returns, with `GetLayout` as the fallback when the category brings none, and the test double hands out one layout for both lookups as the client does (1e07c99).

The header element is also added only when `C_XMLUtil.GetTemplateInfo` knows its template (8981407). The settings list asks for the extent while laying out, and an unknown template is an error that takes the whole page down; that happened in game with an install whose TOC lacked `SettingsHeader.xml`. Test seen to fail with the guard removed.

13804 Lua assertions, 0 failures.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -29,6 +29,7 @@ globals = {
 
     -- WoW API namespaces
     "C_AddOns",
+    "C_XMLUtil",
     "C_Map",
     "C_Container",
     "C_Spell",
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -87,6 +87,7 @@ globals = {
     "PlaySound",
     "Screenshot",
     "hooksecurefunc",
+    "CreateSettingsListSectionHeaderInitializer",
     "seterrorhandler",
     "geterrorhandler",
     "debugstack",
--- a/QuickRoute/Modules/SettingsPanel.lua
+++ b/QuickRoute/Modules/SettingsPanel.lua
@@ -97,9 +97,16 @@ end
 local function RegisterNativeSettings()
     L = QR.L
 
-    local category = Settings.RegisterVerticalLayoutCategory(
+    -- The layout is the second return value, and it is what anything other
+    -- than a checkbox, slider or dropdown has to be added to. Those three take
+    -- the category and register themselves; a section header does not, so it
+    -- needs AddInitializer or it is simply built and dropped -- which is what
+    -- used to happen here, leaving ten translations of SETTINGS_GENERAL with
+    -- nowhere to appear.
+    local category, layout = Settings.RegisterVerticalLayoutCategory(
         L["ADDON_TITLE"] or "QuickRoute"
     )
+    SettingsPanel.layout = layout
 
     -- The header element ("C2" on the design canvas) comes first; the global
     -- SettingsPanel is Blizzard's frame, not this module.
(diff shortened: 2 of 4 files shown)


CybotTM:
Review note (author; Copilot's review quota is exhausted this session). Read the diff at 8981407: the captured second return value, the section header through `CreateSettingsListSectionHeaderInitializer`, the shared layout with the header element from #48 and its `GetLayout` fallback, the `GetTemplateInfo` guard, the double's shared layout and `C_XMLUtil`, and the two tests; both seen to fail with their defect injected.

CybotTM:
Self-review: 89814078c5dad9899ce65268b0b9b096a39efdec

PR #47, opened 2026-09-02
fix: zone-locked beacons are not ready everywhere, and rows keep their names

CybotTM:
Two defects seen in-game on 2026-09-02, both reproduced first in the UI simulator against this addon.

**Zone-locked beacons counted as ready everywhere.** The Kirin Tor Beacon and the Sunreaver Beacon only work on Isle of Thunder and inside the Throne of Thunder, the Mobile Telemancy Beacon only in Suramar; the game refuses them elsewhere. The data knew that only as a free-text `restriction` that nothing read, so in Stormwind all three showed READY, sat in the quick-teleport list and were offered in the map sidebar as a way onto their map. They now carry `usableOnMaps` (504 for both beacons: the Throne of Thunder floors 508–515 are children of Isle of Thunder in the client's UiMap table; 680 for Suramar), `PlayerInfo:IsOnAnyMap` walks the player's map up its parents, and an owned teleport the game would refuse here reads NOT HERE (new status ZONE), is hidden by the "usable" filter, and is left out of the quick list and the sidebar.

**Sidebar and list rows lost their names.** `SetTextToFit()` was called right after `SetText`; the client reads the missing argument as `""` (SimpleFontStringAPIDocumentation, `Default = ""`) and clears the text. The four calls pass the text now. The test double had the method as a no-op, which is why the existing row test stayed green; it now behaves like the client, and three existing `CreateTeleportRow` tests go red on the old call.

Tests: 13720 pass. The new ones were each verified by defect injection: the sidebar call without a text fails one, the panel call fails three, removing the zone checks fails three.

The same commit is on `feat/teleport-cards` ([#43](https://github.com/CybotTM/wow-quickroute/pull/43)), where the sidebar defect was photographed.


CybotTM:
Review note (author, no bot review available: the Copilot request on this head was accepted with an empty `requested_reviewers` read-back, the quota tell). Read the whole diff at ccc3a6b: the `usableOnMaps` data for 95567/95568/140324 against the UiMap ids verified earlier, `PlayerInfo:IsOnAnyMap` including its parent walk and the open-fail when the map API cannot answer, the zone check placed after ownership in both panels, the four `SetTextToFit(text)` sites, and the mock's changed `SetTextToFit`. One cosmetic leftover, a doubled blank line after the zone check in both panels; left as is.

CybotTM:
Self-review: ccc3a6b09bb388f4a5da63e4fe536955cfb00e93

Issue #52, opened 2026-09-03
Quick teleports: long item names are cut at the column boundary


CybotTM:
In the quick-teleport list, item names longer than the name column are cut at the column boundary with no ellipsis and no tooltip fallback: `Beginner's Guide to D`, `Delver's Mana-Boun`, `Personal Key to the A`.

Visible in the quick-teleport screenshot added by [#50](https://github.com/CybotTM/wow-quickroute/pull/50).

**Reproduce:** own any teleport whose item name exceeds the name column, then open the quick-teleport panel from the minimap button.

The destination column beside it is short for most rows, so the split is fixed rather than proportional. Options are a proportional split, an ellipsis plus tooltip, or a wider panel.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_014LbhshbQ8jBjLU2ymEZHbr)_

PR #53, opened 2026-09-03
fix: labels that do not fit are shortened, not cut mid-word

### LINKED ISSUE #51: Teleport panel: filter dropdown label is cut to "Usable Now, Gr"
The teleport panel's filter dropdown shows `Usable Now, Gr` when the availability filter is `usable` and destination grouping is on. The label is `Usable Now, Grouped`; the button is too narrow for it, and the text is cut mid-word rather than shortened or wrapped.

Visible in the teleport panel screenshot added by [#50](https://github.com/CybotTM/wow-quickroute/pull/50), top left of the panel.

**Reproduce:** open the teleport panel with `availabilityFilter = "usable"` and `groupByDestination = true`.

**Worth deciding rather than just widening:** the label combines two independent settings, so its length grows with the filter names. Either the button sizes to its longest label, or the two settings stop sharing one control.

CybotTM:
Closes [#51](https://github.com/CybotTM/wow-quickroute/issues/51) and [#52](https://github.com/CybotTM/wow-quickroute/issues/52). Both are the same shape: a label longer than the space it was given, cut mid-word with nothing to say that anything was removed.

## The filter dropdown

The label is the selected availability filter plus the grouping checkbox, so its length follows the localized strings, while `WowStyle1DropdownTemplate` gives the button a fixed width. `Usable Now, Group by Destination` arrived as `Usable Now, Gr`, and German is longer still.

The button is now sized to the widest label its menu can produce, measured from the strings themselves rather than a constant, and clamped to the panel. `TeleportPanel.FilterLabelCandidates` builds that set — each filter alone and each filter combined with grouping — so a new filter or a new locale is covered without another width being guessed.

## The quick-teleport list

The name and destination columns were 100 px each whatever the panel's width. Together with the icon, the status column and the gaps that is wider than the row, so both columns were short of the space actually available and the name paid for it.

The columns are derived from the row width now, and the panel is 380 px rather than 310. A name that still does not fit — `Beginner's Guide to Dimensional Rifting` will not fit any panel that belongs under the minimap — ends in an ellipsis, cut on a character boundary so a multi-byte character is never split. The row's tooltip already carried the full name; the ellipsis is what tells the reader it is there. The empty-state line is a message rather than a column value, so it spans the label area instead of sitting in the name column.

## Tests

Eight new tests, and each was checked against a mutation that only it catches:

| Mutation | Test that reddened |
|---|---|
| filter and grouping never combined into one label | filter labels include each filter combined with grouping |
| no room reserved for the chevron | the dropdown is wide enough for its longest label |
| minimum width dropped | a short label still gets a usable dropdown width |
| probe text left behind | measuring leaves no text behind on the probe |
| ellipsis not applied | a name too wide for its column ends in an ellipsis |
| UTF-8 continuation back-off removed | shortening never cuts inside a multi-byte character |
| zero-width guard dropped | an unsized label is left alone |

`lua tests/run_tests.lua`: 13909 passed, 0 failed.

Both panels were also rendered in [wow-ui-sim](https://github.com/Osso/wow-ui-sim) against this branch: the dropdown shows `Usable Now, Group by Destination` in full, and the list shows `Beginner's Guide to Dime...` beside names that fit unshortened.

## The screenshots

[#50](https://github.com/CybotTM/wow-quickroute/pull/50) merged before these fixes, so the quick-teleport and teleport-panel images on `main` show the clipped label and the narrow panel. Both are re-rendered here, in a second commit, along with the crop bounds in the render recipe.


CybotTM:
**Self-review** — a Copilot review was requested and the read-back came back empty, which is this repository's quota tell, so no bot review is in flight to wait for. Reviewed against `main` at `977922d`, head `051766b7`.

**`TeleportPanel.lua`** — the gap arithmetic behind `DROPDOWN_CHROME` is a constant rather than a measurement, and `DROPDOWN_MAX_WIDTH` means a locale whose label exceeds 320 px would clip again. That is deliberate: the dropdown sits left of a right-anchored Refresh button in an 820 px panel, and an unbounded width would eventually reach it. The measuring FontString is created once per panel and parented to the dropdown, so it is not a growing allocation. If `dropdown.Text` is ever absent the probe falls back to `GameFontHighlightSmall`, which could measure slightly off; the rendered check below shows it does not on the real template.

**`MiniTeleportPanel.lua`** — three things I checked rather than assumed:

- `GAPS = 8` has to match the actual anchors, not the comment. It does: icon `LEFT +2`, name `LEFT of icon RIGHT +4`, status `RIGHT -2`; the name-to-destination gap is `COLUMN_GAP` and subtracted separately.
- Rows are pooled, and the empty-state line widens `nameLabel` to span the label area. Every other writer — the teleport loop and the mount row — sets it back to `NAME_WIDTH`, so a pooled row cannot inherit the wide column. `destLabel` is never re-sized, so it has no such hazard.
- The shortening loop terminates when even the ellipsis does not fit: `bytes` reaches 0, the label is set to `...`, and the outer condition ends the loop. No unbounded case.

**Verified, not assumed** — both panels rendered in [wow-ui-sim](https://github.com/Osso/wow-ui-sim) against this branch: the dropdown shows `Usable Now, Group by Destination` whole, and the list shows `Beginner's Guide to Dime...` beside names that fit unshortened. `lua tests/run_tests.lua`: 13909 passed, 0 failed. Each of the eight new tests was checked against a mutation only it catches; one of them — the dropdown width — passed a mutation at first because it never asserted the room for the chevron, and was sharpened until it failed.

CybotTM:
Self-review: 051766b737d37264c8d35db032dbd65c94e32ac8

PR #54, opened 2026-09-03
chore(release): 1.16.0

CybotTM:
Release 1.16.0. Fifty-nine commits since [v1.15.0](https://github.com/CybotTM/wow-quickroute/releases/tag/v1.15.0), eleven of them features, so the minor digit moves.

## What is in it

**Added** — the settings page opens with its designed header, and `/qrsurvey` and `/qrdiag` record what only a running game can answer.

**Changed** — the grouped teleport view is drawn as picture cards, and each card is pictured by its destination's zone map rather than by the teleport item.

**Fixed** — thirty-two, the ones a player notices being: zone-locked beacons no longer read as ready everywhere; sidebar and list rows show their names again; the housing teleport works; the teleport list no longer empties itself after a teleport; worn cloaks, rings and slippers count as owned; flight masters are filed on the map the player is actually on; and labels that do not fit are shortened rather than cut mid-word.

## Version surfaces

`QuickRoute/QuickRoute.toc` is the only one — `QR.version` and the settings header both read it through `C_AddOns.GetAddOnMetadata`, so there is no second place to keep in step. Grep confirms no `1.15.0` remains outside the changelog's history.

## Changelog

Two housekeeping changes beyond moving `Unreleased` to `1.16.0`: the block carried two `### Added` headings, which are folded into one, and the label fixes merged as [#53](https://github.com/CybotTM/wow-quickroute/pull/53) had no entry and get one.

`lua tests/run_tests.lua` on this branch: 13909 passed, 0 failed.

After this merges, the tag goes on `main`'s merge commit and CI publishes from it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## [Unreleased]
 
+## [1.16.0] - 2026-09-03
+
 ### Added
+- `/qrsurvey` records what the client says about each map you stand on, and `/qrdiag` records Lua errors and teleport-list rebuilds. Both keep their results across sessions, for the questions that can only be answered from a running game.
 - The settings page opens with the header from the design ("C2"): the logo, the addon's name, a subtitle, a line with version and author and the site below it, drawn over a travel network with one route in gold; below it a status bar that says whether TomTom was found (and which version), how many teleports the character has, and a button that opens the route window. The settings themselves are unchanged, and Blizzard's own Defaults button stays where it is.
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CybotTM:
**Self-review** — a Copilot review was requested on [#53](https://github.com/CybotTM/wow-quickroute/pull/53) an hour ago and the read-back came back empty, which is this account's quota tell, so there is no bot review in flight to wait for. Reviewed against `main` at `138747e`, head `2de9bab2`.

Two files, and the things worth checking on a release bump rather than assuming:

- **Version surfaces.** `QuickRoute/QuickRoute.toc` is the only one. `QR.version` (`QuickRoute.lua:13`) and the settings header (`SettingsHeader.lua:59`) both read it through `C_AddOns.GetAddOnMetadata`, so nothing else carries the number. `grep -rn '1\.15\.0' --include='*.toc' --include='*.lua'` returns nothing after the bump.
- **The version choice.** Fifty-nine commits since v1.15.0, of which eleven are `feat` — minor, not patch.
- **The changelog move is a move, not a rewrite.** Every entry that was under `[Unreleased]` is under `[1.16.0]`, none dropped: the diff adds the `[1.16.0]` heading and relocates the stray `### Added` block, and the only new prose is the entry for #53, which had none.
- **The duplicate heading.** The block carried two `### Added` sections, so the release would have rendered two Added lists. Folded into one, order preserved.

`lua tests/run_tests.lua` on this branch: 13909 passed, 0 failed. CI: 4 of 4.

The release skill's `validate-pre-release.sh` reports three failures, none introduced here: the empty `[Unreleased]` section is what a cut release branch is supposed to look like, the on-main check belongs to the tagging phase, and the missing `id-token: write` / `attestations: write` permissions are a standing gap in `release.yml` that predates this release — filed separately rather than changed inside a release PR.

PR #56, opened 2026-09-04
feat(ci): attest the build provenance of the release package

### LINKED ISSUE #55: Release workflow publishes without build provenance
The release job in `.github/workflows/release.yml` declares `permissions: contents: write` and nothing else, so it cannot produce build provenance for what it publishes. `validate-pre-release.sh` from the release skill reports it on every release:

```
Release infrastructure:
  FAIL  Release workflow permissions (missing: id-token:write, attestations:write)
```

The addon ships to CurseForge and Wago from that job. Nothing today lets someone who downloads the zip check that it was built by this workflow from this tag rather than assembled elsewhere.

**What closing this needs — as a starting point, not a spec:** `id-token: write` and `attestations: write` on the release job, plus `actions/attest-build-provenance` over the published archive. Whether the attestation is worth the extra job time for a WoW addon, and whether CurseForge and Wago surface it to anyone at all, is the part worth deciding before building it — an addon whose consumers cannot verify the attestation gains a file nobody reads.

Noticed while cutting [1.16.0](https://github.com/CybotTM/wow-quickroute/pull/54); not changed inside a release PR.

CybotTM:
Closes [#55](https://github.com/CybotTM/wow-quickroute/issues/55).

The release job publishes `QuickRoute-<version>.zip` with nothing that ties it to this workflow, this repository or this commit. It now attests the package, and the statement covers the exact bytes that get uploaded: the step sits between packaging and the release upload, so it describes the file that is published rather than a rebuild of it.

A downloader checks it with:

```
gh attestation verify QuickRoute-1.17.0.zip --repo CybotTM/wow-quickroute
```

`id-token: write` and `attestations: write` are added to the release job because the OIDC token is what identifies the workflow to the attestation store; without both, the step records nothing. `contents: write` is unchanged.

## What this does not cover

CurseForge and Wago carry no attestation of their own, so this reaches the GitHub download and nothing else. That was the open question in the issue, and it is worth stating plainly rather than leaving implied: most people install this addon through the CurseForge app or WoWUp, and none of them will ever see this. The GitHub asset is the one path where the check is available, and the cost is one step in a job that already runs.

## Verification, and what is not verified

The pinned action was checked rather than assumed: `4d10147` exists in `actions/attest-build-provenance`, its commit date sits five minutes before the `v4.2.2` release, and `subject-path` is a real input of `action.yml` at that SHA. The workflow parses, and the release job's step order puts the attestation between `Package addon` and `Create GitHub release`.

Attestations need a public repository or Advanced Security, and a step that cannot record one fails the job it is in — which would have turned every future release red. This repository is public, so the precondition holds.

The step itself has not run. `release.yml` triggers on `v*` tags only, so the first execution is the next release — deliberately not worked around by adding `workflow_dispatch`, since that job holds `CF_API_KEY` and `WAGO_API_TOKEN` and a manual trigger is a bigger change than this fix. The check to run after the next tag is the `gh attestation verify` line above; if it fails, the fix is wrong and this needs reopening.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,6 +64,11 @@ jobs:
     needs: [tests, luacheck]
     permissions:
       contents: write
+      # Build provenance. The OIDC token is what identifies this workflow, this
+      # repository and this commit to the attestation store; without both of
+      # these the attest step cannot record anything.
+      id-token: write
+      attestations: write
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -157,6 +162,17 @@ jobs:
           cd dist
           zip -r "QuickRoute-${VERSION}.zip" QuickRoute/
           echo "Created QuickRoute-${VERSION}.zip ($(du -h "QuickRoute-${VERSION}.zip" | cut -f1))"
+
+      # Before the upload, so the statement exists for the exact bytes that are
+      # published rather than for a rebuild of them. A downloader checks it with
+      #   gh attestation verify QuickRoute-<version>.zip --repo CybotTM/wow-quickroute
+      # which answers whether this workflow, at this commit, produced that file.
+      # CurseForge and Wago carry no attestation of their own, so this covers the
+      # GitHub download only.
+      - name: Attest the build provenance of the package
+        uses: actions/attest-build-provenance@4d101475d8b20a2381f78447822ac1eab6504dd8 # v4.2.2
+        with:
+          subject-path: dist/QuickRoute-${{ steps.version.outputs.version }}.zip
 
       - name: Create GitHub release
         env:


CybotTM:
**Self-review** — Copilot answered on this head with an error row rather than a review: "unable to review … the user who requested the review has reached their quota limit". That is the quota wall stated outright, not inferred, so there is no review in flight to wait for. Reviewed against `main` at `d5439c1`, head `e6c01d9b`.

Sixteen lines of workflow, and four things checked rather than assumed:

- **The pinned action.** `4d10147` exists in `actions/attest-build-provenance`; its commit is dated five minutes before the [v4.2.2](https://github.com/actions/attest-build-provenance/releases/tag/v4.2.2) release, and `subject-path` is a real input of `action.yml` *at that SHA* — not at `main`, which is the version this workflow will never run.
- **Permission scoping.** A job-level `permissions:` block replaces the default set entirely, so adding two scopes cannot silently drop a third. Parsed: `tests` and `luacheck` keep `contents: read`, `release` now holds `contents: write, id-token: write, attestations: write`. `contents: write` is what `gh release create`/`upload` needs and it is unchanged; there is no top-level block to interact with.
- **Step order.** Parsed from the YAML, not read by eye: `Package addon` → `Attest the build provenance of the package` → `Create GitHub release`. The statement therefore covers the file that is uploaded rather than a rebuild of it.
- **The precondition.** Attestations require a public repository or Advanced Security, and a step that cannot record one fails the job it sits in — which would have turned every future release red. `gh repo view` reports `visibility=PUBLIC`.

**What is not verified, and cannot be from here:** the step has never run. `release.yml` triggers on `v*` tags only, so its first execution is the next release. Adding `workflow_dispatch` would have made it testable and was deliberately not done — that job holds `CF_API_KEY` and `WAGO_API_TOKEN`, and a manual trigger is a larger change than the fix it would be testing. The check after the next tag is `gh attestation verify QuickRoute-<version>.zip --repo CybotTM/wow-quickroute`; if it fails, this is wrong and [#55](https://github.com/CybotTM/wow-quickroute/issues/55) needs reopening.

Repository CI: 4 of 4 green. No test covers workflow files here, so the parse above is the evidence for the YAML.

CybotTM:
Self-review: e6c01d9b23f3b7230aa4bd1269d4683047ec60b4

Issue #66, opened 2026-09-08
While moving, the addon spikes every ~second and stutters the game


Playnot:
### What happened?

Whenever I'm stationary, everything seems fine, but when I start moving, every one second or so, the game stutters for a few ms.
I noticed in the addons page there are CPU spikes. Not present when the addon is disabled. I also disabled TomTom, in case the integration is at fault, didn't help.

https://imgur.com/uatm5im

### Expected behavior

The game shouldn't stutter, that was very high usage.

### Steps to reproduce

1. Add addon
2. Enable
3. Start walking
4. Game stutters every ~second

### Debug info

```markdown

```

### Lua errors (if any)

```text

```

### Additional context

Additional addons currently enabled:

<details>
<summary>Show enabled addons (96)</summary>

```text
AllTheThings
AstralKeys
Auctionator
BetterFishing
BetterWardrobe
BetterWardrobe_SourceData
BigWigs
BigWigs_Core
BigWigs_KhazAlgar
BigWigs_LiberationOfUndermine
BigWigs_NerubarPalace
BigWigs_Options
BigWigs_Plugins
!BugGrabber
BugSack
CanIMogIt
SharedMedia_Causese
ChattyLittleNpc
Clique
Details
DoesMyAltKnowThat
GTFO
HandyNotes
HandyNotes_WarWithin
HandyNotes_DragonGlyphs
Hekili
Leatrix_Plus
LiteMount
LittleWigs
LittleWigs_BattleForAzeroth
LittleWigs_BurningCrusade
LittleWigs_Cataclysm
LittleWigs_Classic
LittleWigs_Dragonflight
LittleWigs_Legion
LittleWigs_MistsOfPandaria
LittleWigs_Shadowlands
LittleWigs_WarlordsOfDraenor
LittleWigs_WrathOfTheLichKing
MapPinTimers
MRT
MythicDungeonTools
Narcissus
Narcissus_Achievements
Narcissus_BagFilter
Narcissus_Barbershop
Narcissus_Database_Item
Narcissus_Database_NPC
Narcissus_GamePad
OPie
OribosExchange
ParagonReputation
Pawn
PepeGoldTracker
tdBattlePetScript
PetTracker
PetTracker_Battle
PetTracker_Config
PetTracker_Journal
Plumber
PremadeGroupsFilter
RaiderIO_DB_EU_M
RaiderIO
RaiderIO_DB_EU_R
RareScanner
Rematch
SimpleItemLevel
[...]

PR #67, opened 2026-09-08
fix: stop re-routing every tracked quest each second, and go idle in combat

CybotTM:
Fixes the stutter reported in #66, and stops the addon working through fights.

## What was happening

`QuestTeleportButtons` picks each tracked quest's teleport out of a computed route, and caches that per quest. The cache could never be hit: the event handler wiped it wholesale on every event it listens to, and `QUEST_LOG_UPDATE` is one of them. That event fires about once a second while the player moves, so each firing routed every tracked quest again from scratch — one Dijkstra run over a 278-node, 3237-edge graph per quest. The position bucket in the cache key was a thousandth of the map, a few yards, so anything that survived the wipe was invalidated by walking anyway.

That matches the report exactly: nothing while standing still, a hitch about once a second while moving, no QuickRoute window involved. It also explains why the cost grew as more addons were installed — the expensive path hangs off events any addon can provoke by asking the client about quests, spells or bags, and each firing paid full price.

Separately, none of this stopped for combat, and neither did the rest of the addon.

## Measurements

`scripts/benchmark_movement.lua`, the repository's own movement benchmark — 25 tracked quests, QuickRoute window closed, ten simulated seconds at 60 fps. `de47ddf` versus this branch, same host. Standalone Lua CPU, not in-client FPS.

| workload | before | after |
|---|---|---|
| moving | 226 routes, 227.8 ms, worst frame 3.70 ms | 50 routes, 54.4 ms, worst frame 1.66 ms |
| stationary, one quest event per second | 249 routes, 199.8 ms, worst frame 4.70 ms | **0 routes, 0.74 ms**, worst frame 0.02 ms |
| stationary, settled | 0 routes, 0.20 ms | 0 routes, 0.16 ms |
| initial warmup | 25 routes, 13.9 ms | 25 routes, 13.9 ms |

The second row is the reported bug: a quest event a second, going nowhere, previously cost ten routes a second. The first row keeps the routes that a real change of position earns — two bucket crossings across the fixture's path, times 25 quests.

The earlier movement work recorded in `docs/PERFORMANCE-REVIEW-2026-09-06.md` made each route cheaper and left their number alone ("226 in both"); this reduces the number.

Combat, measured separately by event with the same mock:

| in combat | before | after |
|---|---|---|
| `SUPER_TRACKING_CHANGED` | 40 ms, graph rebuild + route | nothing |
| map pin moved | 2.2 ms, route | nothing |
| bag / spell / toy / equipment / skill events | a full inventory scan each | nothing |
| zone crossing | survey record written | nothing |
| a fight of 20 events, 4 tracked quests | cache emptied, all 4 re-routed after | cache kept, none re-routed |
| on leaving combat | — | one scan, one route, once |

## What changed

**Recompute budget.** The cache wipe is limited to the events that change which quests are tracked, plus a cooldown change the module has already confirmed. Everything else a cached entry depends on — position bucket, the graph it was computed against, whether a rebuild is pending, the entry's age, the teleport's cooldown — is checked on every read, so those events lost no freshness by not wiping. `SPELLS_CHANGED` and `BAG_UPDATE_DELAYED` invalidate themselves through the graph identity the entries record. The position bucket became a named constant at 20 per axis; `CACHE_TTL` is what bounds staleness, at 30 seconds, and did so at the old divisor too.

Combat also stopped emptying the quest cache on every event: the handler returns without doing button work either way, so the wipe only moved a full recompute of every tracked quest to the moment the fight ended, next to the deferred scan and the graph rebuild. A change to which quests are tracked still invalidates mid-fight.

**Combat.** Three gates, deferring rather than dropping. A waypoint change is remembered and resumed by the leave-combat callback — the flag records *which* of the two paths is owed, because resuming a cleared pin through the change path would pop the route window open after any fight in which the player removed their pin. The inventory rescan leaves its pending flag set so a fight's events fold into one scan afterwards, carrying the accumulated forced-rebuild flag; if the leave-combat callback never arrives (a reload or logout mid-fight), the next inventory event releases the flags so the scan is not suppressed for the rest of the session. The zone survey is the exception and drops its record, because a capture run after the fight would describe wherever the player ended up — the invented evidence that recorder exists to avoid.

## Tests

`tests/test_route_recompute_budget.lua` and `tests/test_combat_idle.lua`. Every guard was watched to fail with its fix reverted. Two of those mutations were informative rather than confirming: one made the position bucket blind to the mapID and passed, which is why the same-coordinates-different-zone case exists; another showed the first cleared-waypoint test to be a no-op, because it re-implemented the dispatch instead of calling it — extracting `ResumeDeferredWaypointWork` from the callback closure is what let the test exercise the real thing.

Suite: 16028 → 16060 assertions, 0 failures, in both orders; luacheck clean.

One existing expectation changed deliberately. `test_inventory_graph_invalidation.lua` pinned "combat scans the bags but postpones the rebuild"; combat now postpones the scan as well, so it asserts the new contract — nothing during the fight, the change picked up in a single scan afterwards, still exactly one route invalidation.

## The trade-off this makes

A quest whose objective moves while the player stands still — an escort stepping into the next zone, say — can keep its previous teleport button for up to `CACHE_TTL`, 30 seconds, where before a quest event replaced it within the debounce. That bound is not new: `WaypointIntegration`'s own coordinate cache holds each quest's resolved position for the same 30 seconds, so re-routing on every quest event was recomputing from coordinates that were themselves up to 30 seconds old. Adding the resolved waypoint to the cache key would compare one 30-second-old value against another and buy nothing.

## Not addressed here

`PlayerInventory.lua:599` forces a graph rebuild for every event that is not `BAG_UPDATE`, bypassing the teleport-set comparison beside it. `SPELLS_CHANGED` therefore still costs about 75 ms out of combat: a full graph rebuild plus a route per tracked quest. Narrowing it needs a cheap capability signature to compare against — riding and flying are spells, so a naive narrowing would miss a player learning to fly — and that is a change to `TravelRequirements`, not to this fix.

I have not been able to confirm this is what the reporter of #66 hit; their version and settings are still unknown. The mechanism matches the video in every respect I could check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+- Stopped recalculating every tracked quest's route about once a second while moving, which stuttered the game and grew more expensive the more addons were installed. Routes are now reused until the player has actually moved a meaningful distance, changed zone, or changed which quests are tracked.
+- Left the game alone during combat. Routes, inventory scans and zone survey records are no longer computed mid-fight; the postponed work runs once the fight is over, except a zone crossing, which is dropped rather than recorded late against the wrong position.
+
 ## [1.18.1] - 2026-09-07
 
 ### Fixed
--- a/QuickRoute/Modules/PlayerInventory.lua
+++ b/QuickRoute/Modules/PlayerInventory.lua
@@ -593,6 +593,50 @@ eventFrame:RegisterEvent("TOYS_UPDATED")
 eventFrame:RegisterEvent("SPELLS_CHANGED")
 eventFrame:RegisterEvent("SKILL_LINES_CHANGED")
 
+--- The scan the debounce timer runs. Also called from the leave-combat callback
+-- below, for the case where the timer landed mid-fight and postponed it.
+function PlayerInventory:RunDeferredScan()
+    PlayerInventory.pendingScan = false
+    PlayerInventory.scanDeferredByCombat = false
+    local force = forceGraphRefresh
+    forceGraphRefresh = false
+
+    -- Skip scan if addon not fully initialized yet
+    if not QR.db then return end
+
+    -- Perform the scan
+    local before = not force and PlayerInventory:GetAllTeleports()
+    PlayerInventory:ScanAll()
+    local changed = force or not SameTeleports(before, PlayerInventory:GetAllTeleports())
+
+    -- Notify PathCalculator if it exists (defer during combat to avoid expensive graph rebuild)
+    if changed and QR.PathCalculator and QR.PathCalculator.OnInventoryChanged then
+        if InCombatLockdown() then
+            QR.PathCalculator.graphDirty = true
+        else
+            QR.PathCalculator:OnInventoryChanged()
+        end
+    end
+
+    if QR.debugMode then
+        local count = PlayerInventory:GetTeleportCount()
+        QR:Debug(string_format("Inventory updated (%d teleports)", count))
+    end
+end
+
+--- Register the leave-combat callback. Called from Initialize rather than file
+-- scope: QR:RegisterCombatCallback lives in QuickRoute.lua, which the .toc
+-- loads after this file.
+function PlayerInventory:RegisterCombatCallback()
+    if self.combatCallbackRegistered or not QR.RegisterCombatCallback then return end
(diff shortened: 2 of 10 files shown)


CybotTM:
Self-review: c042ff6bb41d7b8e3f34e787bf5ce876ee210c91

PR #69, opened 2026-09-08
perf: stop recomputing routes and re-asking the client, and key the quest cache on its destination

CybotTM:
Three reviews of 1.18.2 — CPU hot paths, memory and allocation, and client API call counts — each required to measure rather than infer. This fixes what they found that is both worth fixing and safe to fix, and says plainly what it leaves.

## What was wrong, and what it costs now

**Re-routing every quest when the arrow moves.** `SUPER_TRACKING_CHANGED` emptied the quest route cache, and 1.18.2 recomputed every tracked quest from it. Super-tracking decides which quest carries the arrow, not what any quest's route is — `WaypointIntegration:GetQuestWaypoint` takes an explicit questID and never consults `C_SuperTrack`. Measured with 25 tracked quests: all 25 routes recomputed, none of the 25 values different. It fires on every turn-in and accept, on tracker and map clicks, and whenever another addon calls `SetSuperTrackedQuestID`. `QUEST_WATCH_LIST_CHANGED` wiped the same cache for the one quest that changed, where `PruneQuestCache` already dropped what stopped being watched.

| event, 25 tracked quests | before | after |
|---|---|---|
| `SUPER_TRACKING_CHANGED` | 17.7 ms, 25 routes | 0.09 ms, 0 routes |
| `QUEST_WATCH_LIST_CHANGED` | 18.7 ms, 25 routes | 0.13 ms, 0 routes |

**Cooldowns asked once per route instead of once per refresh.** Every route prices every teleport against its cooldown, and a refresh computes one route per quest on its own frame, so the same question reached the client once per quest: 188 client queries for one refresh of 25 quests on a 52-teleport character, 2300 for 92 teleports on a fuller one. `GetCooldown` now memoizes inside a refresh batch — 188 → 66.

**Answers that cannot change, asked repeatedly.** Zone display names (906 → 443 `GetMapInfo` over ten refreshes), quest titles fetched ahead of the cache that made them unnecessary, and spell cast times re-read per priced teleport. Total client calls over ten refreshes: 11,667 → 10,766.

**Allocation and layout at 5 Hz.** The button positioning handler built two tables and two closures per tick (2.9 KiB, ~50 MiB an hour standing still) and re-anchored every button whether or not its block had moved. Tables and closure reused; re-anchoring happens when the block moves.

## A correctness fix that came out of reviewing this

An adversarial review of the branch found that the route cache key had no field for the quest's **destination**. Completing an objective can advance a quest to one in another zone without the player moving a step, so position, graph and age all still matched and the entry went on offering the teleport for where the quest used to point — a button that teleports to the wrong continent. Removing `SUPER_TRACKING_CHANGED` took away the accident that had been covering this for the super-tracked quest; the other 24 were never covered, because `QUEST_LOG_UPDATE` did not invalidate either. The destination is now part of the key, which closes it for all of them.

Three smaller ones from the same review: the readiness check that consumes `SPELL_UPDATE_COOLDOWN` could read the batch memo it exists to check, and now closes the batch first; ending a batch is generation-aware, so a superseded refresh cannot close its successor's; an abandoned batch expires after two seconds rather than freezing cooldowns addon-wide until the next refresh. A quest title that had not streamed in yet is no longer frozen into the coordinate cache as "Quest Objective".

The readiness-check fix ships without a guard. Two attempts to pin it passed for the wrong reason or could not tell the fixed and unfixed code apart — the reviewer reproduced it by stepping an async batch frame by frame, and the mock runs `C_Timer.After` synchronously. Said plainly rather than dressed up in a test that proves nothing.

## What this deliberately does not do

- **`IsQuestFlaggedCompleted`, 1705 calls for 5 distinct ids.** `TravelRequirements` evaluates access requirements live at search time on purpose; its own comment says so and two tests pin it, including one that an unavailable quest API stays *unknown* rather than becoming a remembered false. A memo trades a documented guarantee for call count. I tried it: four tests failed, correctly.
- **`ReconnectPlayerNode`, ~95 KiB per call, up to 1.26 MiB/s while moving.** The largest remaining allocation source. The CPU review's ablation puts an upper bound of 5–11 % of route time on any fix here, and its first simulated version came out slower. Worth doing on its own, with its own measurements, not folded into this.
- **The multi-stop tour matrix, count² cells at ~386 KiB each** (151 MiB transient at 20 stops). Halving it rests on route symmetry, which is an assumption about the graph rather than a fact I measured.
- **`forceGraphRefresh` on every non-`BAG_UPDATE` inventory event**, still ~141 ms per `SPELLS_CHANGED` on a stocked character. Narrowing it needs a capability signature that cannot miss a newly learned flying skill — riding and flying are spells — which is a `TravelRequirements` change.
- **The teleport panel's duplicate spell lookups** (766 client calls per list rebuild, roughly 2× per row). A window-open cost, and merging name and icon means restructuring row construction.
- **`ZoneSurvey` at its 3000-record cap** (~3 MiB of SavedVariables) is an intentional bound, and the unused node index and `RemoveNode`'s reverse scan measured 3 % and 7.8 % of a route respectively.

Figures in `docs/PERFORMANCE-REVIEW-2026-09-06.md` for the moving benchmark ("226 in both") are superseded by 1.18.2 and again here; that document is left as the dated record it is.

## Tests

16,072 → 16,080 assertions, 0 failures, both orders; luacheck clean. Every guard was watched to fail with its fix reverted.

Two gaps the reviews exposed are now covered rather than argued about. The positioning handler had no test at all — the mock ships no `ObjectiveTrackerFrame`, so `OnUpdate` returned immediately in every existing assertion — and now has three, including the recycled-button case that catches a stale remembered anchor. The cooldown memo has one for the aliasing trap: these queries return module-level tables that the next query overwrites, so a memo of references answers every id with whatever was asked last. The test as first written compared a spell against an item, which are different tables even with the bug; it compares two spells now and fails when the memo stores references.

### DIFF
--- a/QuickRoute/Core/PathCalculator.lua
+++ b/QuickRoute/Core/PathCalculator.lua
@@ -147,10 +147,20 @@ end)
 -- @param nodeName string The graph node name (may contain English zone names)
 -- @param mapID number The map ID for the node
 -- @return string The localized display name
+-- A zone's name does not change while the player is logged in, and every step
+-- of every route asks for one: measured at 1430 identical queries a minute for
+-- two distinct maps. Keyed by node and map because the answer also depends on
+-- the node's own parenthetical.
+local localizedNodeNames = {}
+
 local function GetLocalizedNodeDisplayName(nodeName, mapID)
     if not mapID or not C_Map or not C_Map.GetMapInfo then
         return nodeName
     end
+
+    local memoKey = nodeName .. "\0" .. mapID
+    local remembered = localizedNodeNames[memoKey]
+    if remembered then return remembered end
 
     local mapInfo = C_Map.GetMapInfo(mapID)
     if not mapInfo or not mapInfo.name then
@@ -160,16 +170,18 @@ local function GetLocalizedNodeDisplayName(nodeName, mapID)
     local zoneName = mapInfo.name
 
     -- Check if node name has a parenthetical disambiguation (e.g., "Dalaran (Broken Isles)")
+    local display = zoneName
     local _, parenthetical = nodeName:match("^(.+)%s*%((.+)%)$")
     if parenthetical and mapInfo.parentMapID then
         -- Get the localized parent (continent/region) name
         local parentInfo = C_Map.GetMapInfo(mapInfo.parentMapID)
         if parentInfo and parentInfo.name and parentInfo.name ~= zoneName then
-            return zoneName .. " (" .. parentInfo.name .. ")"
+            display = zoneName .. " (" .. parentInfo.name .. ")"
         end
     end
 
-    return zoneName
+    localizedNodeNames[memoKey] = display
+    return display
 end
 
 -------------------------------------------------------------------------------
--- a/QuickRoute/Core/TravelTime.lua
+++ b/QuickRoute/Core/TravelTime.lua
@@ -303,9 +303,20 @@ function TravelTime:GetTeleportTime(teleportData, teleportID, sourceType)
         if ok and Number(itemSpellID) and itemSpellID > 0 then spellID = itemSpellID end
     end
     if spellID and C_Spell and C_Spell.GetSpellInfo then
-        local ok, info = pcall(C_Spell.GetSpellInfo, spellID)
-        if ok and type(info) == "table" and Number(info.castTime) then
-            castTime = info.castTime / 1000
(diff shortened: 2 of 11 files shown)


CybotTM:
Self-review: 255b00613143f158072fc7190e2ffccd5a75e4e9

PR #74, opened 2026-09-09
fix: stop rebuilding the whole travel graph on every SPELLS_CHANGED

CybotTM:
The reporter of #66 tried 1.18.3 and says the stutter is still there — every five to ten seconds now rather than every second, still only while moving. Their new recording shows client-wide CPU sitting at 3–4 % and spiking to 25–53 % while walking.

## What it is

Measured on 1.18.3 with 51 teleports and 25 tracked quests, one firing of each event the inventory watcher listens to:

| event | cost | rebuilds | routes |
|---|---|---|---|
| `QUEST_LOG_UPDATE` | 0.95 ms | 0 | 0 |
| **`SPELLS_CHANGED`** | **44.38 ms** | **1** | **8** |
| `TOYS_UPDATED`, `PLAYER_EQUIPMENT_CHANGED`, `BAG_UPDATE`, `SKILL_LINES_CHANGED` | ~0.06 ms | 0 | 0 |

44 ms in a single frame is a visible stutter at any frame rate, and `SPELLS_CHANGED` fires constantly for reasons that have nothing to do with travel — which is also why the cost rose with the number of addons installed. The graph build alone is 32–36 ms on this character; the rest is a route for each quest that then finds a new graph.

The 1.18.3 work removed the per-second cost, which is why the reported cadence moved from "every second" to "every five to ten seconds". This is what is left.

## The fix

Every event except `BAG_UPDATE` marked the graph dirty, on the grounds that it might have changed something the teleport comparison cannot see. What that protects is now stated rather than assumed:

- the teleport set is already compared entry by entry, and an equipped teleport carries its slot in its entry, so equipment reaches that comparison;
- `PlayerInfo:CapabilitySignature` covers the rest of what `CanUseTeleport` reads — faction, class, race, Engineering — from four values the module already caches.

The requirement checks the build makes for flight edges depend on quest and phase state. The blanket force never covered those either — nothing fires `SPELLS_CHANGED` when a quest is completed — and `TravelRequirements` continues to mark the graph dirty on the zone and world events that do change them.

Same fixture after: `SPELLS_CHANGED` costs **0.29 ms**, no rebuild, no routes.

## Tests

Four tests pinned the blanket force, asserting that these events invalidate even when nothing changed. Their subject — named in the production comment they guard — is that a capability change is not lost when the event coalesces into a pending `BAG_UPDATE` batch, the "looting while learning a riding spell" case. They now do that literally, by giving the character Engineering, and gain the half that was not testable before: an event that changed nothing invalidates nothing.

Dropping the signature fails eight assertions; restoring the blanket force fails four. Suite 16,084 → 16,088, 0 failures in both orders; luacheck clean.

Not confirmed against the reporter's client yet — this is measured here, and the cadence and the addon-count dependence both match what they describe.

### DIFF
--- a/QuickRoute/Modules/PlayerInventory.lua
+++ b/QuickRoute/Modules/PlayerInventory.lua
@@ -598,8 +598,13 @@ eventFrame:RegisterEvent("SKILL_LINES_CHANGED")
 function PlayerInventory:RunDeferredScan()
     PlayerInventory.pendingScan = false
     PlayerInventory.scanDeferredByCombat = false
-    local force = forceGraphRefresh
+    -- What the graph was built against, besides the teleports themselves. A
+    -- change here has to rebuild; an unchanged signature means the comparison
+    -- below is the whole story.
+    local capabilities = QR.PlayerInfo and QR.PlayerInfo:CapabilitySignature()
+    local force = forceGraphRefresh or (capabilities ~= PlayerInventory.graphCapabilities)
     forceGraphRefresh = false
+    PlayerInventory.graphCapabilities = capabilities
 
     -- Skip scan if addon not fully initialized yet
     if not QR.db then return end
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review: 2f18d16e97f456741792b5528250ba3e99845931

Issue #72, opened 2026-09-09
Cooldown batch: the readiness check strips a running refresh, and a waypoint gap disarms a good button


CybotTM:
Two smaller items from the adversarial review of #69, kept together because both are about the batch that #69 introduced and neither is worth a release on its own.

**The readiness check strips a running refresh of its memo.** `UpdateCooldownState` closes the cooldown batch as its first act, which is right — it has to see the client rather than the batch's remembered answers, or it reports "nothing changed" about the teleport the player just used. But `SPELL_UPDATE_COOLDOWN` fires on every global cooldown, and when the check then finds nothing moved it returns without cancelling, so the refresh carries on for its remaining frames with no memo:

```
undisturbed refresh of 12 quests:                          3 client cooldown reads
same, with one unrelated SPELL_UPDATE_COOLDOWN mid-batch:  12
before #69 (no batch at all):                             10
```

Correctness is unaffected — live reads are strictly more correct — but it costs #69 its own saving on the path that fires most often. The shape of a fix is to read past the memo instead of closing the batch. An attempt to do that by adding a live-read accessor broke 25 existing tests, which change cooldown state mid-batch and expect the change to be seen, so it needs more care than a one-line swap.

**A transient waypoint gap now evicts a good entry.** Since #69 resolves the waypoint on every read, a refresh during which the client cannot say where a quest points produces a nil destination, misses the cache, deletes the entry and clears the button's action:

```
before #69:  entry intact, tooltip "Stormwind",      1 route calculation
after  #69:  entry gone,   tooltip "Calculating...", 2 route calculations
```

The player sees a button that is on screen but inert, and after `PENDING_GRACE` (2 seconds) of a continued gap it hides. Before, a cache hit never consulted the waypoint layer, so the button rode the gap out untouched. `PENDING_GRACE`'s own comment — "retain an inactive icon across a brief data gap" — is the evidence that these gaps happen.

**Also worth doing, and cheap:** nothing in the suite can catch either of these, or the batch lifecycle generally, because the mock runs `C_Timer.After` synchronously. The review built a guard that is red before #69's fix and green after by giving the mock's CooldownTracker stand-in a real `BeginBatch`/`EndBatch` and a memo, then firing `SPELL_UPDATE_COOLDOWN` between two queued refresh callbacks. Lifting that into `tests/` would close the blind spot.

PR #75, opened 2026-09-09
fix: compare where a quest points, not just which map

### LINKED ISSUE #70: Quest route cache compares the destination map, so an objective moving within one zone keeps the old teleport
`GetCachedTeleportForQuest` keys the quest route cache on the destination **map** (`QuestTeleportButtons.lua`, `cached.destMapID == destMapID`). An objective that advances to another point in the *same* zone therefore hits the cache and keeps the previous teleport.

Measured, objective moving from `x=0.1` to `x=0.9` in one zone, everything else unchanged:

| | first step offered |
|---|---|
| before #69 | correct — the blanket `SUPER_TRACKING_CHANGED` wipe caught it |
| after #69 | stale — `destMapID` unchanged, so the entry is reused |

It is not hypothetical. `PathCalculator.lua` wires a destination into the graph by its coordinates (`ConnectNearbyNodes(destName, destMapID, destX, destY)`), so two points in one zone attach to different nearby nodes and can select a different first teleport. Large zones and capitals with more than one landing point are where it shows.

Bounded: `CACHE_TTL` is 30 seconds, and it clears as soon as the player moves far enough to change the position bucket.

Map-only was chosen deliberately and that reasoning still holds — `C_QuestLog.GetNextWaypoint` walks a multi-step quest along its path, so x and y drift while the destination does not, and comparing them would recompute routes that cannot have changed. Closing this needs something between the two: a coarse bucket on the destination coordinates, or comparing the graph node the destination resolves to rather than the map.

Found by an adversarial review of #69, which cleared that PR to merge and ranked this as a follow-up.

CybotTM:
Closes #70 and the testable half of #72, both recorded while reviewing #69.

## The destination is compared at the right resolution

#69 keyed the quest route cache on the destination **map**. That missed an objective advancing to another part of the same zone: the graph wires a destination in by its coordinates (`ConnectNearbyNodes(destName, destMapID, destX, destY)`), so two points in one zone attach to different nearby nodes and can select a different first teleport. Before #69 the blanket wipe caught that case, so it was a regression against the release before it.

Comparing exact coordinates is not the answer either — `C_QuestLog.GetNextWaypoint` walks a multi-step quest along its path, so they drift while the destination stays put, and every drift would recompute a route that cannot have changed.

The destination is now bucketed on the same grid that decides the player has moved. Both halves are pinned: an objective crossing its zone is routed again, one drifting along its path is not. Coarsening the key back to the map fails the first test; comparing exact coordinates fails the second.

## The cooldown batch is testable now

The async harness's `CooldownTracker` stand-in was `{GetCooldown = function() return {ready = true} end}` — no memo, so no assertion routed through it could tell a remembered cooldown from a fresh one. That is why the readiness-check fix in #69 shipped without a guard after two attempts that passed for the wrong reason.

The stand-in now behaves like the real module: it remembers answers for the duration of a batch and counts client reads. The guard exists and fails when the readiness check reads an open batch instead of the client — which is the defect #69 fixed blind.

## One test-hygiene fix found on the way

`WaypointIntegration`'s coordinate cache outlives a test, so a test that moved an objective left the next one resolving the old position. The fixture clears it. That is what made the drift test fail for a reason that had nothing to do with the code under test.

## Verification

16,088 → 16,092 assertions, 0 failures in both orders; luacheck clean. Every guard was watched to fail with its fix reverted.

### DIFF
--- a/QuickRoute/Modules/QuestTeleportButtons.lua
+++ b/QuickRoute/Modules/QuestTeleportButtons.lua
@@ -99,6 +99,22 @@ local INVALIDATING_EVENTS = {
 -- what bounds how stale this choice can get, and it did so at 1000 too.
 local POSITION_BUCKETS = 20
 
+--- Where a quest points, at the same resolution as the player's own position.
+-- Keyed on the map alone this missed an objective advancing to another part of
+-- the same zone -- a destination is wired into the graph by its coordinates, so
+-- two points in one zone attach to different nearby nodes and can pick a
+-- different first teleport. Keyed on exact coordinates it would recompute
+-- constantly instead: C_QuestLog.GetNextWaypoint walks a multi-step quest along
+-- its path, so the coordinates drift while the destination does not. The same
+-- grid that decides the player has moved decides the objective has.
+local function DestinationBucket(waypoint)
+    if not waypoint or not waypoint.mapID then return nil end
+    local x, y = waypoint.x, waypoint.y
+    if type(x) ~= "number" or type(y) ~= "number" then return waypoint.mapID .. ":?" end
+    return string_format("%d:%d:%d", waypoint.mapID,
+        math_floor(x * POSITION_BUCKETS), math_floor(y * POSITION_BUCKETS))
+end
+
 local function GetPositionBucket()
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review: 7d3a4410153f5dc36eb6b17e37a5f20b460b5dc3

Issue #71, opened 2026-09-09
Resolving the quest waypoint on every cache read is cold about once a second while moving


CybotTM:
#69 made the quest route cache compare the destination, which means `ResolveQuestWaypoint` now runs on every read, including cache hits. The Dijkstra-per-quest saving is fully retained — an all-hit refresh of 25 quests still performs **0** `CalculatePath` calls — but the waypoint resolution is new work on that path:

```
before #69, all-hit refresh:  0 GetQuestWaypoint calls
after  #69, all-hit refresh:  8 GetQuestWaypoint calls
```

The comment justifying it is wrong about what governs the cost. It says the resolution is cheap because `WaypointIntegration` caches coordinates for 30 seconds. That TTL is not what decides: `WaypointIntegration` wipes `questCoordCache` wholesale on `QUEST_LOG_UPDATE` and `QUEST_POI_UPDATE`, and that fires about once a second while the player moves. So while moving the resolution is cold on essentially every refresh.

A cold resolution measured **9 client API calls** for the easy case, where the waypoint is found on the player's own map. The expensive case — the broad zone scan that `GetQuestWaypoint`'s negative caching exists for — was not measured and needs an in-game number: 25 quests times that cost, about once a second while moving, is the same shape as the regression #66 was about.

Sub-note: on the hit path the resolution still passes `ignoreNegativeCache` for a quest whose button is pending, so those reads bypass the negative cache and always take the full scan.

Worth measuring in a live client before deciding the shape of a fix.

Found by an adversarial review of #69, which cleared that PR to merge and ranked this as a follow-up.

PR #77, opened 2026-09-09
fix: the global cooldown is not a reason to replan every quest

CybotTM:
The reporter of [#66](https://github.com/CybotTM/wow-quickroute/issues/66) says 1.18.4 did not settle it, and gave the two observations that find it: the spikes track distance and his movement ability, and *"I also noticed spikes whenever I use pretty much any ability."*

## Any ability press re-routed every tracked quest

Items and toys share the global cooldown, so pressing any ability at all makes every item and toy teleport report about 1.5 seconds of unavailability. `UpdateCooldownState` took that for a change in readiness, and a readiness change empties the quest route cache and re-routes every tracked quest.

Measured with 52 teleports and 25 quests tracked, per keypress:

| | cost | routes | client cooldown reads |
|---|---|---|---|
| before | 13.82 ms | 8 | 118 |
| after | **0.02 ms** | **0** | 52 |

No route can change over a window that closes before the player could act on the answer, so a teleport that is one global cooldown away from usable now counts as ready **for the purpose of deciding whether to replan**.

`CooldownTracker` deliberately refuses that inference when it *reports* a cooldown — a short real cooldown is a real cooldown, and its comment says so explicitly. That is the reporting question; this is the planning one, and they want different answers. A real teleport cooldown still invalidates and still re-routes.

The remaining 52 reads per press are the detector walking the teleport list. That is C-side, unchanged here, and tracked in [#71](https://github.com/CybotTM/wow-quickroute/issues/71)'s neighbourhood.

## Still open after this

The other half of the report — *"it seems to be distance-based, more often when I use my movement ability"* — is the position bucket: crossing one re-routes every tracked quest, and a movement ability crosses buckets quickly. That is a burst of eight routes at 2–3 ms each, one per frame, and it is not addressed here.

## Tests

Both directions pinned in the file that already owns the global-cooldown rules, each watched to fail: treating the global cooldown as a change again re-routes on the keypress, and treating everything as ready loses a real cooldown.

The first version of the guard asserted against a suite with no tracked quests, so it passed whether or not the code was right — it sets up its own quests now. Suite 16,092 → 16,095, 0 failures in both orders; luacheck clean.

Measured here, not in the reporter's client. His description is what located it.

### DIFF
--- a/QuickRoute/Modules/QuestTeleportButtons.lua
+++ b/QuickRoute/Modules/QuestTeleportButtons.lua
@@ -97,6 +97,10 @@ local INVALIDATING_EVENTS = {
 -- to walk to -- is not answered by any divisor: CACHE_TTL already expires every
 -- entry within 30 seconds regardless of where the player stands, so that is
 -- what bounds how stale this choice can get, and it did so at 1000 too.
+-- The longest a teleport can be unavailable without it being worth replanning.
+-- The global cooldown is 1.5 seconds and every item and toy teleport shares it.
+local GLOBAL_COOLDOWN = 1.5
+
 local POSITION_BUCKETS = 20
 
 --- Where a quest points, at the same resolution as the player's own position.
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review: 3622300c4e56aa83aea096f07cc2ccf28bed00d8

PR #78, opened 2026-09-09
fix: measure how far the player moved, not which cell they are in

CybotTM:
The other half of what the reporter of [#66](https://github.com/CybotTM/wow-quickroute/issues/66) described — *"it seems to be distance-based, more often when I use my movement ability"*. [#77](https://github.com/CybotTM/wow-quickroute/pull/77) took the ability half.

## A grid has edges, and players work along them

Whether a cached route still applied was decided by comparing grid cells: the player's position rounded to a twentieth of the map. Anyone circling an objective, strafing in a fight, or blinking back and forth across one of those edges re-routed every tracked quest each time they crossed it.

Measured with 25 tracked quests, ten steps of 0.0002 of the map back and forth:

| | route calculations |
|---|---|
| across a cell boundary | **80** |
| the same steps a little to either side | 0 |
| across a boundary, after this change | **0** |

The question a cache asks is whether the answer it holds is still the answer, and that is a distance from where it was computed — which has no edges to flap across. The repository's own movement benchmark halves as a side effect: 50 route calculations and 64 ms over ten simulated seconds become 25 and 25.6 ms, because the fixture's path no longer straddles a boundary.

The bucket stays where it belongs. `OnMovementUpdate` uses it to decide *when to look again*, which is a different question from *whether the answer changed*, and a cell is a fine answer to it.

## Tests

Stepping back and forth across one spot never re-routes; the existing cases for a real move across the zone and for an objective drifting along its path still hold. Restoring the cell comparison fails the new test with 30 calculations.

Suite 16,095 → 16,096, 0 failures in both orders; luacheck clean.

Measured here rather than in the reporter's client, as before — his description of *where* the spikes cluster is what located this one.

### DIFF
--- a/QuickRoute/Modules/QuestTeleportButtons.lua
+++ b/QuickRoute/Modules/QuestTeleportButtons.lua
@@ -103,6 +103,10 @@ local GLOBAL_COOLDOWN = 1.5
 
 local POSITION_BUCKETS = 20
 
+-- How far the player may move before a computed route is asked for again, as a
+-- fraction of the map. The same span as one position bucket, without the edges.
+local POSITION_TOLERANCE = 1 / POSITION_BUCKETS
+
 --- Where a quest points, at the same resolution as the player's own position.
 -- Keyed on the map alone this missed an objective advancing to another part of
 -- the same zone -- a destination is wired into the graph by its coordinates, so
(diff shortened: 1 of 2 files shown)


CybotTM:
Self-review: 221e3d54abbd51aa5b5b40ad69c9f4a9f8d19549

PR #80, opened 2026-09-09
fix(quest-buttons): keep the cooldown memo and the route across a transient event

CybotTM:
Both halves of [#72](https://github.com/CybotTM/wow-quickroute/issues/72), from the adversarial review of [#69](https://github.com/CybotTM/wow-quickroute/pull/69).

## The readiness check stripped a running refresh of its memo

`UpdateCooldownState` ended the cooldown batch as its first act so it would see the client rather than the batch's remembered answers. Seeing live is right; ending is more than that costs. `SPELL_UPDATE_COOLDOWN` fires on every global cooldown and almost always finds nothing moved, so a refresh that was still running went on for its remaining frames with no memo.

| refresh of 12 quests | client cooldown reads |
|---|---|
| undisturbed | 3 |
| one unrelated `SPELL_UPDATE_COOLDOWN` mid-batch | 12 |
| before #69, no batch at all | 10 |

`CooldownTracker` gains `SuspendBatch`/`ResumeBatch`. The scan reads live either way — that part is unchanged. What differs is afterwards: the batch is ended only when a readiness really moved, which is also the moment its memo stops being true, and otherwise it is reopened with what it had.

The issue records an earlier attempt at this that broke 25 tests by adding a live-read accessor. This does not need one: nothing reads past the memo, the memo is simply not discarded.

**What resuming the batch stops detecting:** the readiness check counts anything within a global cooldown of usable as ready, so a teleport going from 1.4 seconds remaining to genuinely ready is not a change — and the resumed memo still answers "not ready" for it. The button for that teleport appears one refresh later than it would have. The window is bounded by the batch, which is a frame per tracked quest and at most `BATCH_MAX_SECONDS`. That is the price of the 12-reads-to-3 above.

## A momentary waypoint gap evicted a good entry

Since [#69](https://github.com/CybotTM/wow-quickroute/pull/69) resolves the waypoint on every read, a frame in which the client cannot say where a quest points produced a nil destination. That read as a *changed* destination, missed the cache, recomputed with no waypoint, came back incomplete and deleted the entry — an icon on screen that does nothing until `PENDING_GRACE` hides it. A nil now means "unknown", not "changed".

**What this stops detecting:** a quest that genuinely loses its waypoint for good keeps its cached teleport until `CACHE_TTL` expires it, the player moves out of tolerance, or the graph changes. That is what happened before #69, bounded at 30 seconds.

## The harness could not have caught either

The `CooldownTracker` stand-in in `test_questbutton_async.lua` answered with `ready` but no `remaining`, and the readiness check counts anything within a global cooldown of usable as ready. So `ready = false` still read as ready, every flip of `state.cooldownReady` in that file was invisible to the check, and no guard there could have depended on one. Fixed in the same commit — it is what makes the new guards mean anything.

## Tests

Both watched to fail with their fix reverted: the batch guard reports 5 client reads instead of 3 for the two quests queued after the event, and the waypoint guard finds the cache entry gone and the button's action cleared.

Suite 16,096 → 16,103, 0 failures; luacheck 0 warnings across 122 files.

### DIFF
--- a/QuickRoute/Modules/CooldownTracker.lua
+++ b/QuickRoute/Modules/CooldownTracker.lua
@@ -285,6 +285,29 @@ function CooldownTracker:EndBatch()
     if self.batchMemo then wipe(self.batchMemo) end
 end
 
+--- Read the client rather than the memo, without discarding it.
+-- For the one caller that has to see live state in the middle of somebody
+-- else's batch: the readiness check that decides whether a cooldown moved. It
+-- used to end the batch outright, which is correct but throws away the memo of
+-- a refresh that is still running -- and SPELL_UPDATE_COOLDOWN fires on every
+-- global cooldown, so that was the common case rather than the rare one.
+-- @return boolean Whether a batch was open, to hand back to ResumeBatch
+function CooldownTracker:SuspendBatch()
+    local wasOpen = self.batchOpen or false
+    self.batchOpen = false
+    return wasOpen
+end
+
+--- Reopen a batch suspended by SuspendBatch, with its memo intact.
+-- Only sound when the caller established that nothing it cares about moved --
+-- otherwise the memo would answer the rest of the batch with the state from
+-- before the change. EndBatch is the other exit, and it is the right one
+-- whenever readiness did move.
+-- @param wasOpen boolean The value SuspendBatch returned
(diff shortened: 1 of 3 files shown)


CybotTM:
Self-review: a5b42418cdb209452eb664ba1d447d974aaf951e

PR #81, opened 2026-09-09
feat(waypoints): measure what resolving each tracked quest's waypoint costs

CybotTM:
[#71](https://github.com/CybotTM/wow-quickroute/issues/71) is blocked on one number, and it is a number this repository cannot produce: what `GetQuestWaypoint`'s broad zone scan costs. The easy case — the waypoint found on the player's own map — was measured at 9 client API calls. The expensive case needs a live client with real quests. `/qrwpcost` produces it.

## What it does

For every tracked quest: clear the coordinate cache, time one cold resolution, then time one warm one. It reports both, plus where the quest resolved — the player's own map, another map, or not at all.

The shape of the report, with invented numbers — nothing here has been measured yet, that is the point of the PR:

```
quest     cold ms  warm ms  resolved
84721       0.412    0.008  map 2274 (the player's own)
84903      11.740    0.009  map 2339
...

cold total : 38.221 ms for 25 quest(s), worst single 11.740 ms
warm total : 0.204 ms
resolved   : 6 on the player's own map, 18 elsewhere, 1 not found
```

The cold total is the one that matters: it is what a refresh pays whenever `questCoordCache` has just been wiped, and `WaypointIntegration` wipes it wholesale on `QUEST_LOG_UPDATE` and `QUEST_POI_UPDATE` — about once a second while the player moves. Compare it against one frame at your frame rate.

## Two decisions worth stating

**Milliseconds, not a count of client calls.** Counting would mean wrapping `C_QuestLog` functions, and an addon that overwrites a field of a `C_` table taints Blizzard's own quest tracker until the next reload. Time is also the number [#71](https://github.com/CybotTM/wow-quickroute/issues/71) actually asks for.

**Refused in combat, not deferred.** A measurement that runs once the fight is over describes different quests in a different place — the same reasoning `ZoneSurvey` uses for dropping a capture rather than postponing it.

**It empties the coordinate cache.** Measuring the cold case means clearing `questCoordCache`, so a run leaves it empty for every quest but the last, and the next refresh pays the cold cost once. A quest update does the same thing about once a second, so this is benign — but the report says it, because a diagnostic that discards state it does not own should.

## The mock had no clock

`debugprofilestop` was missing from `tests/mock_wow_api.lua`. It now advances by a configurable step on every call, so a measured interval is a predictable non-zero number instead of always 0, and a test can assert on a total.

## Tests

Registration; the report over two tracked quests, one on the player's map and one elsewhere; and an empty tracker, which is reported rather than measured. The warm pass is pinned as a genuine cache hit by counting the client resolution it starts from — clearing the cache before it as well takes that from 2 to 4 and fails the test.

Suite 16,096 → 16,106, 0 failures; luacheck 0 warnings across 122 files.


CybotTM:
Self-review: d85b6d8f78b799725b4ac4e83f9d1d72626d5031

PR #86, opened 2026-09-12
Resolve known hearthstone inns before first use

CybotTM:
Existing hearthstone bindings were omitted from routes until QuickRoute observed a rebind or a successful arrival. Resolve known inns directly from the client's localized bind name, so their approximate destination is available before first use. Observed character bindings always take precedence.

Morgenluft defaults to the inn in its current Midnight version when no matching observation is available. Both localized area-name aliases resolve to the sourced modern innkeeper position. The tooltip identifies this default and its approximate position. The existing observer remains active: a successful arrival or new binding replaces the default, including a binding in the legacy area. Catalogue inference never writes a default as a saved observation.

The compact catalogue contains 59 distinct destinations on 44 maps: 60 coordinate-bearing alias records and 59 ambiguity/phase blockers. Names and collisions were audited against the complete retail 12.1.0.69587 AreaTable in all 11 supported locales. The index is cached, published only after complete localization, and retried at world-entry events if the API was unavailable. Other ambiguous names remain unresolved; the explicit Morgenluft aliases do not bypass additional conflicting candidates.

Coordinates approximate inns, and the catalogue does not cover every inn. Per-location sources, exclusions and a reproducible offline audit are included.

Validation:
- 18,402 Lua assertions passed in both discovery and reverse order; 56 Python tests passed.
- Luacheck: zero warnings/errors across 131 files.
- Complete source/collision audit passed in all 11 locales. Morgenluft coordinates match the current NPC mapper data.
- Regression tests cover real routing with a warm graph, observed-binding precedence, ambiguity, unavailable APIs, reentrant hooks, sparse tables, and the actual tooltip handler.
- Production event-handler tests verify that hearthstone arrivals and new bindings on both legacy and modern maps override the default, invalidate routes, and cancel stale capture callbacks.
- Independent code and data review completed. Controlled warm-lookup measurements show no retained heap growth or repeated localization calls. Live-client FPS and gameplay are not measured by this harness.


CybotTM:
Reviewed at `092655d0`, the current head, in a detached worktree. One finding, and it is not a blocker.

**Gates, re-run here rather than read from the description:** 18402 Lua assertions pass in discovery order and in reverse; luacheck reports 0 warnings and 0 errors over 131 files. Those match what the description claims.

`scripts/audit_hearthstone_catalog.py` was **not** run. It needs the eleven AreaTable CSV exports, the UiMap export, the pinned guide archive and the Wowhead page in a source directory, none of which are available here. The locale and collision audit is therefore unverified from my side, and it is the part of this change I would most want a second pair of eyes on, because every ambiguity decision rests on it.

**Finding: the `ambiguous` flag is not pinned by anything, and it is currently redundant.**

Removing `inn.ambiguous or` from the guard in `BuildInnIndex` leaves all 18402 assertions passing. The reason is in the data rather than in the code: all 59 ambiguous records carry no coordinates, so `ResolveMapPosition` returns nil and the `not mapID` branch already forces `index[name] = false`. The flag and the missing coordinates say the same thing today.

That makes it a latent gap rather than a present defect. The moment somebody adds coordinates to an ambiguous record — to document where the place is while keeping it unroutable, which is a reasonable thing to want — the guard stops doing anything and the record silently becomes routable. A test that builds an index from one ambiguous record **with** coordinates and asserts the name still resolves to nothing would close it, and would have failed under the mutation above.

**What I checked and found correctly pinned:** observation precedence. Moving the catalogue lookup ahead of the saved-binding branch immediately fails `Hearthstone catalog: saved HEARTHSTONE_BOUND on map 94 overrides the modern default` on four assertions, including the one that the result is no longer a catalogue approximation. The distinction between an inferred position and an observed one survives into `ResolveTeleport` as `isApproximate` and `hearthstoneDefault`, which is what the description promises.

**One thing to confirm on a live client rather than here:** the retry path. An index that could not be completed is rebuilt only on `PLAYER_ENTERING_WORLD`. If `C_Map.GetAreaInfo` is cold for a catalogued area during a long session with no zone change, the index stays empty until the next world entry. Worth watching once in practice; the standalone harness cannot show it.

Nothing else in the diff changed my reading of it. `BuildInnIndex` setting `graphDirty` while it runs inside `BuildGraph` is harmless, because `BuildGraph` assigns `graphDirty` from its own success at the end, after `AddPlayerTeleportEdges` has run.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Ajsba9TsThQzrPKu1RxL5W)_

CybotTM:
Pushed `773d29e` with the test the finding above asks for, rather than leaving it as a note: it builds an ambiguous record that does carry coordinates and asserts the name still resolves to nothing. Removing `inn.ambiguous or` from `BuildInnIndex` turns it red, which the whole suite did not before.

18404 assertions pass in discovery and reverse order; luacheck is clean over 131 files. Nothing else in the branch was touched.

Correction to my own finding while I am here: the data invariant was already pinned. `tests/test_hearthstone_catalog.lua` asserts that a blocked area carries no `mapID`, so the two guards now cover each other rather than one standing alone.

`scripts/audit_hearthstone_catalog.py` remains unrun here for want of the client-data exports, and the retail cases are still open.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Ajsba9TsThQzrPKu1RxL5W)_

CybotTM:
Round 3, at `4de3d42`. No new defect. It was run in the main session rather than by an independent reviewer: two attempts at a third review agent died on the account's session limit without producing a report, and repeating an expensive run a third time was not worth it. That is a weaker review than rounds 1 and 2, and it is the reason this note exists.

**Both markers, both directions.** Six mutations against the two exported flags, each built clean and run against the full suite: set `destApproximate` unconditionally (reddens "no leg of an observed route is marked as a guess"); never set it (reddens three assertions in the guessed-landing test); set `destDefault` unconditionally (reddens "an ordinary catalogue record does not claim its map was chosen"); never set it (reddens the chosen-map test on the step and on the summary); drop each field from `STEP_FIELDS` and from the summary (one assertion each); and drive `destApproximate` from `hearthstoneDefault` instead of `isApproximate` (six failures). None survived.

**The open item from round 2 is settled by a count.** `PathCalculator.lua:2137` substitutes `0.5, 0.5` for a node that has a map and no position, which would export an invented coordinate with no marker. Instrumented and run: **0 hits across 18,882 assertions**. Nothing the suite drives reaches `BuildSteps` with such a node, so it is a defensive line rather than a live path. It is pre-existing and untouched by this branch; recorded here rather than changed.

**Gates on the merged tree**, run with CI's own command strings: 18,882 Lua assertions in discovery and in reverse order, 56 Python tests via `python3 -m unittest discover -s tests -p 'test_*.py'`, luacheck 0 warnings and 0 errors over 141 files. Working tree verified identical to the commit after every mutation.

**What remains unverified**, unchanged from round 1: `scripts/audit_hearthstone_catalog.py` needs eleven AreaTable exports, a UiMap export and a pinned guide archive that are not available here, so the catalogue's collision-completeness claim rests on that script alone.

_Assisted by claude-code:claude-opus-5 — [Session](https://claude.ai/code/session_01Ajsba9TsThQzrPKu1RxL5W)_

CybotTM:
Self-review: 4de3d42643b5ffccc591391de88779a7af963c9c

This pull request requires a review and no bot review is in flight; a review by the author satisfies that requirement. The diff on this head was reviewed by the PR author; this comment is the on-the-record attestation the merge gate reads back. It stops matching on the next push.

